PR #462, opened 2014-12-21
Accept vimeo https urls

codejunky:
It is possible that a vimeo URL starts is secured with https. It would be awesome if it would be supported.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/462)


soimort:
`any_download` is used internally - Generally it's not a good idea to use it as a stable API, since I'm planning to refactor all the legacy code.

The trick is that we convert all URLs to `http` form (and handle redirects when possible) before passing them to `any_download`, therefore it's not needed to take `https` patterns into consideration inside extractors (personally I see no necessity to download videos via HTTPS, though I know it's arguable).

Please feel free to do so as long as it meets your needs (just be aware that your regexp failed the test because it doesn't match `http://`.)

Thanks for the PR!

PR #464, opened 2014-12-23
new-site: redtube.com

kaction:
[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/464)


soimort:
We do not have plan to provide specific support for a site like this (because such sites are **too many**). Technically, those sites are very similar and easy to handle (they embedded the video source in HTML directly), so I'll prefer a more "generic" extractor for all of em (that's going to be implemented in the future)

Thanks for the PR anyway!

PR #476, opened 2015-01-21
maybe typo

pastebt:
Can not find sina_download_by_id in source, it may typo of sina_download_by_vid

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/476)


soimort:
Nice catch.

PR #498, opened 2015-02-28
Remove global in subclass of VideoExtractor()

mozbugbox:
Using `site = Youku()` as global variable will prevent the extractor
from getting different video links simutaneously. Each extractor
should create its own Youku() instance in order to avoid interference
under multithreading.

### DIFF
--- a/src/you_get/extractors/youku.py
+++ b/src/you_get/extractors/youku.py
@@ -163,9 +163,19 @@ class Youku(VideoExtractor):
             if not self.streams[stream_id]['src'] and self.password_protected:
                 log.e('[Failed] Wrong password.')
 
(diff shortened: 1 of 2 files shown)


zhangn1985:
do you have test?
if download vid1 and vid2, one by one.
youku_download_by_vid(vid1)
youku_download_by_vid(vid2)
what happens when download vid2?

zhangn1985:
we can reset self.url/self.vid when download videos.

zhangn1985:
@mozbugbox try: 
https://github.com/zhangn1985/you-get/commit/3641662cb8d29e103e4b66dd6012c854916e6a3f

Issue #504, opened 2015-03-20
[Support Request]Any plan to download lizhi FM audio?


crystoneme:
Do you have any plan to add support to download audio of [lizhi FM](http://www.lizhi.fm)?

soimort:
Thanks to @jackyzy823, this was fixed in #506

PR #506, opened 2015-03-25
support for lizhifm

jackyzy823:
support for #504 both single audio or playlist
single like :`http://www.lizhi.fm/#/549759/18864883431656710`
playlist like:`http://www.lizhi.fm/#/31365/`
current playlist max download count:65535 ,which is enough for use.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/506)


soimort:
Manually merged (this branch needs a rebase)

PR #514, opened 2015-04-12
Fix facebook extractor

daniellowtw:
Previously would throw error if there is no hd_src
Modified to use json parser and check if hd_src is None

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/514)


soimort:
Sorry for the delay. Thanks for the PR!

Issue #519, opened 2015-05-12
爱奇艺视频无法解析


tyang34:
错误信息如下： 
$ you-get http://www.iqiyi.com/v_19rrnscu0w.html
Traceback (most recent call last):
  File "/usr/bin/you-get", line 9, in <module>
    load_entry_point('you-get==0.3.32', 'console_scripts', 'you-get')()
  File "/usr/lib/python3.4/site-packages/you_get/__main__.py", line 91, in main
    main()
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 1000, in main
    script_main('you-get', any_download, any_download_playlist)
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 893, in script_main
    download_main(download, download_playlist, args, playlist, output_dir=output_dir, merge=merge, info_only=info_only)
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 785, in download_main
    download(url, **kwargs)
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 993, in any_download
    m.download(url, **kwargs)
  File "/usr/lib/python3.4/site-packages/you_get/extractors/iqiyi.py", line 79, in iqiyi_download
    assert info["code"] == "A000000"
AssertionError

tyang34:
问题并没有解决，请帮忙看一看
错误信息：
$ you-get http://www.iqiyi.com/v_19rrnscu0w.html
Traceback (most recent call last):
  File "/usr/bin/you-get", line 9, in <module>
    load_entry_point('you-get==0.3.32', 'console_scripts', 'you-get')()
  File "/usr/lib/python3.4/site-packages/you_get/__main__.py", line 91, in main
    main()
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 1001, in main
    script_main('you-get', any_download, any_download_playlist)
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 893, in script_main
    download_main(download, download_playlist, args, playlist, output_dir=output_dir, merge=merge, info_only=info_only)
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 785, in download_main
    download(url, **kwargs)
  File "/usr/lib/python3.4/site-packages/you_get/common.py", line 994, in any_download
    m.download(url, **kwargs)
  File "/usr/lib/python3.4/site-packages/you_get/extractors/iqiyi.py", line 91, in iqiyi_download
    assert info["code"] == "A000000"
AssertionError

[...]

PR #520, opened 2015-05-12
iqiyi changes its encrypt key

jackyzy823:
Fix for #519 
add changelog in iqiyi.py for others checking iqiyi player swf version.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/520)


soimort:
Awesome.

PR #523, opened 2015-05-15
now get iqiyi enc key automatically

jackyzy823:
iQiyi changes it's swf version frequently,so enc key changes too.
So before downloading,get swf from video page and extract the key.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/523)

### DIFF
--- a/src/you_get/extractors/iqiyi.py
+++ b/src/you_get/extractors/iqiyi.py
@@ -12,6 +12,9 @@ import hashlib
 
 '''
 Changelog:
+-> http://www.iqiyi.com/common/flashplayer/20150514/MainPlayer_5_2_21_c3_2_6_2.swf
+    In this version ,it changes enc key to 'Qakh4T0A'
+    consider to write a function to parse swf and extract this key automatically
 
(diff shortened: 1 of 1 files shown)


PeterDing:
@jackyzy823 @soimort 
I has seen this commit, I alse write a script for [rg3/youtube-dl](https://github.com/rg3/youtube-dl) using your finding algorithms.
I commit to [this](https://github.com/rg3/youtube-dl/compare/master...PeterDing:iqiyi).
So, I ask the permission to create pull request for rg3/youtube-dl.

yan12125:
Note: youtube-dl is released into public domain, so codes to enter it should be released into public domain, too.

jackyzy823:
@PeterDing  feel free to use these codes.

PeterDing:
@jackyzy823 Thank you.

yan12125:
@jackyzy823 iQiyi codes are merged into youtube-dl now. Would you like to be listed as the original author of [iqiyi.py at youtube-dl](https://github.com/rg3/youtube-dl/blob/master/youtube_dl/extractor/iqiyi.py)?

jackyzy823:
it's ok,just under `jackyzy823` . New method is introduced in #543 ,anyone wants to merge this to youtube-dl?

PR #526, opened 2015-05-19
The new player changes enc key

phpengineer:
In this version iqiyi player, it changes enc key

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/526)


jackyzy823:
in #523 , i write a function to get enc key automatically.

soimort:
Thanks for the PR - #523 provided a better solution.

PR #536, opened 2015-06-01
Fix case when hMusic data is None

pini-gh:
This occurs for example with these urls:
http://music.163.com/song?id=17699228
http://music.163.com/song?id=17699232

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/536)


soimort:
Thanks!:+1:

PR #546, opened 2015-06-16
Fix zhanqi

liushuyu:
Zhanqi extractor works again...
And added ts merging implementation...

Signed-off-by: liushuyu <liushuyu_011@163.com>

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/546)

### DIFF
--- a/src/you_get/extractors/zhanqi.py
+++ b/src/you_get/extractors/zhanqi.py
@@ -38,14 +38,15 @@ def zhanqi_download(url, output_dir = '.', merge = True, info_only = False):
             real_url.append(i)
         type_ = ''
         size = 0
-        _, type_, temp = url_info(real_url[0])
-        size = temp*(len(real_url))
-        
+        for url in real_url:
+        _, type_, temp = url_info(url)
+        size += temp or 0
+
         print_info(site_info, title, type_ or 'ts', size)
         if not info_only:
             download_urls(real_url, title, type_ or 'ts', None, output_dir, merge = merge)
     else:
-        NotImplementedError(Unknown_video_type)
+        NotImplementedError('Unknown_video_type')
 site_info = "zhanqi.tv"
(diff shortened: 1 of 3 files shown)


jackyzy823:
some urls for vods return 404 in zhaqi.py Ln.41  such as `http://www.zhanqi.tv/videos/dota2/2015/06/23670.html`.

liushuyu:
Oh ,f**k
Zhanqi updated its server again!!!
Maybe I need to change...

liushuyu:
@jackyzy823 I fixed this again! Why not giving a try?

liushuyu:
Zhanqi will split video to hundred of thousands of video parts, it will be very hard (slow) to get the real size.
And if you make users wait so long, users will think 'Oh, no responding...' 
So I really can't think up a solution...

src/you_get/extractors/zhanqi.py
```diff
@@ -7,22 +7,46 @@
...
+        vod_m3u8 = get_html(vod_m3u8_request)
+        part_url = re.findall(r'(/[^#]+)\.ts',vod_m3u8)
+        real_url = []
+        for i in part_url:
+            i = vod_base + i + ".ts"
+            real_url.append(i)
+        type_ = ''
+        size = 0
+        for url in real_url:
+        _, type_, temp = url_info(url)
```
soimort: Have you actually run this part of code? It's an `IndentationError` here.

liushuyu:
I am so sorry for that.
I had a fever last night, and I can't remember what I have done...

src/you_get/extractors/zhanqi.py
liushuyu: Fixed ....

soimort:
Never mind, everybody makes mistakes.
Thanks for the PR.

PR #591, opened 2015-08-10
new iqiyi key

jackyzy823:
for http://www.iqiyi.com/common/flashplayer/20150805/MainPlayer_5_2_26_c3_3_7.swf

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/591)


jackyzy823:
new swf comes http://www.iqiyi.com/common/flashplayer/20150810/MainPlayer_5_2_26_c3_3_7_1.swf. this key still works. I think it will still work for 4 to 5 days.

PR #605, opened 2015-08-23
Add Weibo Miaopai support(微博秒拍视频)

cnbeining:
Something like this.

I did not use the M3U file, but the mobile version instead.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/605)


jackyzy823:
thanks

PR #497, opened 2015-02-26
Some bilibili has duplicate url

pastebt:
Such as http://www.bilibili.com/video/av2020012/

Before patch, I got:

Video Site: bilibili.com
Title:      【科幻】黑夜传说四部合集（720P修复版）（2003-2012）（双语字幕）
Type:       Flash video (video/x-flv)
Size:       6527.82 MiB (6844915737 Bytes)

Real URLs:
['http://nl-amsterdam.acgvideo.com/c/ac/3124235-1.flv?expires=1424949600&ssig=-kFTiFPQImuQ_2pYh3ibMA&o=1179620804&rate=0', 'http://nl-amsterdam.acgvideo.com/c/ac/3124235-1.flv?expires=1424949600&ssig=-kFTiFPQImuQ_2pYh3ibMA&o=1179620804&rate=0', 'http://ws.acgvideo.com/5/70/3124236-1.flv', 'http://ws.acgvideo.com/2/6f/3124237-1.flv', 'http://ws.acgvideo.com/0/31/3124238-1.flv']

First two urls are same, which waste bandwidth

After patch:

Title:      【科幻】黑夜传说四部合集（720P修复版）（2003-2012）（双语字幕）
Type:       Flash video (video/x-flv)
Size:       4962.75 MiB (5203818377 Bytes)

Real URLs:
['http://nl-amsterdam.acgvideo.com/c/ac/3124235-1.flv?expires=1424951400&ssig=n8baoSz5LQtuGh7BAlxGig&o=1179620804&rate=0', 'http://ws.acgvideo.com/5/70/3124236-1.flv', 'http://ws.acgvideo.com/2/6f/3124237-1.flv', 'http://ws.acgvideo.com/0/31/3124238-1.flv']

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/497)

PR #610, opened 2015-08-27
fix duplicate url bug for bilibili

CzBiX:
当有多 P 时，会错误的将 1P 的 id 重复加入 list。

测试地址：
http://www.bilibili.com/video/av824538/
http://www.bilibili.com/video/av2772797/

fix #497

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/610)


jackyzy823:
thanks

soimort:
I believe this patch introduced a bug #688. Let's revert it.

PR #615, opened 2015-08-30
Support 178.com

zhangn1985:
only youku video source

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/615)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -922,7 +922,7 @@ def script_main(script_name, download, download_playlist = None):
             sys.exit(1)
 
 def url_to_module(url):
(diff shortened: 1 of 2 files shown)


zhangn1985:
try: you-get http://dota.178.com/201508/234900468407.html

current only support http://data.178.com/

soimort:
Thanks for the PR. However, I'm not sure if we should add support for all kinds of trivial sites like this one - it simply embeds videos from Youku. Nothing more.

A better way would be to implement a universal extractor for all sites that embed videos hosted elsewhere. That is on the future plan.

zhangn1985:
@soimort got your idea, I will submit new patch,

PR #618, opened 2015-09-02
Fix problems with videos that do not have 720p mode

ddumitran:
Some videos that use lower resolutions do not work for me.
Here is an example: www.dailymotion.com/video/x2mzuat

The fix is quite simple (arguably, using get() instead of [] inside try-catch might be better).

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/618)


soimort:
:+1:

Issue #635, opened 2015-09-10
UnicodeEncodeError


kupo:
you-get https://www.youtube.com/watch?v=AH61HoM5NI4
site:                YouTube
Traceback (most recent call last):
  File "/usr/local/bin/you-get", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.4/dist-packages/you_get/__main__.py", line 91, in
 main
    main()
  File "/usr/local/lib/python3.4/dist-packages/you_get/common.py", line 1018, in
 main
    script_main('you-get', any_download, any_download_playlist)
  File "/usr/local/lib/python3.4/dist-packages/you_get/common.py", line 909, in
script_main
    download_main(download, download_playlist, args, playlist, output_dir=output
_dir, merge=merge, info_only=info_only)
  File "/usr/local/lib/python3.4/dist-packages/you_get/common.py", line 801, in
download_main
    download(url, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/you_get/common.py", line 1011, in
 any_download
    m.download(url, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/you_get/extractor.py", line 46, i
n download_by_url
    self.download(**kwargs)
  File "/usr/local/lib/python3.4/dist-packages/you_get/extractor.py", line 164,
in download
    self.p(stream_id)
  File "/usr/local/lib/python3.4/dist-packages/you_get/extractor.py", line 109,
in p
    print("title:               %s" % self.title)
UnicodeEncodeError: 'ascii' codec can't encode character '\u300c' in position 33
: ordinal not in range(128)

zhangn1985:
maybe same as #622

soimort:
It seems you're on a Unix-like system. Please set your locale as Unicode-compatible encoding, e.g. `LANG="en_US.UTF-8"` or `LANG="zh_CN.UTF-8"`, otherwise it's not possible to display a video title with CJK character in it.

PR #641, opened 2015-09-12
get encoding from response date

### LINKED ISSUE #622: encode get failed when response header doesn't have charset
when some site response header without charset, get charset will failed.
eg:
print(response.headers)
Server: DnionOS/1.2.1.3
Date: Thu, 03 Sep 2015 09:04:32 GMT
Content-Type: text/html
Content-Length: 28827
Vary: Accept-Encoding
Age: 1433
Via: GDST-248-65 (DLC-3.0), CT-ZJHZ-C-11-186 (DLC-3.0)
Connection: close

and the site is not utf-8 encoding.
then decode failed.

try to get encode info from response data.

zhangn1985:
fix #622

many sites don't have charset in response header.
eg. sohu, letv,

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/641)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -210,7 +210,8 @@ def get_content(url, headers={}, decoded=True):
 
     # Decode the response body
     if decoded:
-        charset = match1(response.getheader('Content-Type'), r'charset=([\w-]+)')
+        charset = match1(response.getheader('Content-Type'), r'charset=([\w-]+)') or \
+                  match1(str(data), r'charset=([\w-]+)')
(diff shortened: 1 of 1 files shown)


soimort:
#635 is unrelated to this, since you can always suppose Unicode (UTF-8) is the right one for YouTube.

Could you please be more specific how encoding issue affect those sites? (like, give some example videos that will throw exceptions on encoding failures.)

zhangn1985:
似乎也没有神码影响

PR #645, opened 2015-09-14
[youku] support multi-page playlists

junzh0u:
Youku playlist can have multiple pages, for example: http://www.youku.com/playlist_show/id_22394567.html

Currently we only deal with the first page. And it's not even possible to download the second page via the complete url, for example: http://www.youku.com/playlist_show/id_22394567_ascending_1_page_2.html
Because we always try to extract the playlist id from the url and then download the first page.

With this pull request, we will download all videos from the playlist.

Tested by running the following command locally and make sure the output has all videos included.
you-get -i http://www.youku.com/playlist_show/id_22394567.html

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/645)


soimort:
:+1:

PR #646, opened 2015-09-14
[common] fix skipping download of file existence in some cases

junzh0u:
Current logic of skipping download when file already exists have some bugs when working together with ffmpeg-merge.

TL;DR: When merging multi parts using ffmpeg, we change the ext name of the output file. Thus messed up with the current skipping download mechanism.

Considering a video with title "TITLE", it have multiple parts, all with ext ".flv".
We will look for "TITLE.flv" first. If it doesn't exist, we will download all parts as "TITLE[00].flv", "TITLE[01].flv"... etc.
Then if we don't have ffmpeg installed, we merged them into "TITLE.flv".

So far so good.
But problem happens if we have ffmpeg installed. We will merge all parts to "TITLE.mp4". That means when we try to download the same video again, it will not be able to find any file named "TITLE.flv". Thus it will download again.

This diff solves this issue.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/646)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -499,6 +499,24 @@ class DummyProgressBar:
     def done(self):
         pass
 
+def get_output_filename(urls, title, ext, output_dir, merge):
+    merged_ext = ext
+    if (len(urls) > 1) and merge:
+        from .processor.ffmpeg import has_ffmpeg_installed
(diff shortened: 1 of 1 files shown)


soimort:
Sounds reasonable. Thanks for the contribution!

PR #651, opened 2015-09-16
[youku] gracefully handle single failure when downloading playlist

junzh0u:
TL;DR: When downloading a playlist, a single failure shouldn't terminate the whole process.

Youku playlist can have broken video links. I believe this happens really rare, but I have one sample here: http://www.youku.com/playlist_show/id_21492543.html
This playlist contains a broken video link: http://v.youku.com/v_show/id_XNjUxNzQyNzEy.html

Since python set is unordered, when downloading this playlist, we can hit this video at a random point of the whole process, and crash.

This diff is just a rough idea. I'm open for opinions. I have several in my mind actually:
Should we have a flag guarding this feature?
Should we apply this feature to all extractors? If yes is there a better way to do that other than changing every extractor that support playlist?

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/651)

### DIFF
--- a/src/you_get/extractors/youku.py
+++ b/src/you_get/extractors/youku.py
@@ -6,6 +6,7 @@ from ..extractor import VideoExtractor
 
 import base64
 import time
+import traceback
 
 class Youku(VideoExtractor):
     name = "优酷 (Youku)"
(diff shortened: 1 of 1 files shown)


jackyzy823:
thanks @junzh0u .
New ideas are welcomed!

PR #686, opened 2015-10-05
Load Mozilla cookies.sqlite

soimort:
The old-fashioned Netscape `cookies.txt` is nowhere to be seen in modern browsers. We should really move on to a modern format (e.g. Mozilla's SQLite-based cookies), so there will be no need for one to manually export `cookies.txt` anymore.

Also, this will allow for easier downloads of Bilibili's member-only videos, just grab a cookie file from your Firefox:

```
$ you-get -i -c ~/.mozilla/firefox/gggfz9xl.default/cookies.sqlite \
  http://www.bilibili.com/video/av2987444
```

Bilibili uses CAPTCHA verification as part of its login process, so there is no way to automatize this like we do in Niconico. Preloading existed cookies is a must in such occasions.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/686)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -8,6 +8,7 @@ import platform
 import re
 import sys
 from urllib import request, parse
+from http import cookiejar
 
 from .version import __version__
 from .util import log
@@ -19,6 +20,7 @@ json_output = False
 force = False
 player = None
 extractor_proxy = None
+cookies = None
 cookies_txt = None
 
 fake_headers = {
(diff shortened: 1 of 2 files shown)


soimort:
The plan is to support the old plaintext `cookies.txt`, Mozilla `cookies.splite` and Chrome `Default/Cookies`. I'm not merging this til we're fully done.

soimort:
Implementing Chromium cookies is non-trivial job - the `encrypted_value` stored in SQLite was encrypted using AES (therefore we need a 3rd party crypto library like [pycrypto](https://pypi.python.org/pypi/pycrypto) to handle the algorithmic thing).

Even more sophisticated, based on this article http://n8henrie.com/2013/11/use-chromes-cookies-for-easier-downloading-with-python-requests/, the encryption methods differ on Linux and OS X (didn't mention Windows).

Forget about Chrome, fair enough to support `moz_cookies` and `cookies.txt` only.
(To my knowledge, not only Firefox, but also some WebKit based browsers such as GNOME Web, makes use of `moz_cookies` too)

PR #691, opened 2015-10-07
[sina] match vid from the page

lilydjwg:
fix for http://video.sina.com.cn/view/249851143.html

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/691)


soimort:
:ok_hand:

PR #703, opened 2015-10-16
Revert "fix duplicate url bug for bilibili"

soimort:
Reverts soimort/you-get#610, as it caused a regression bug #688. To reproduce:

```
λ  you-get-develop [develop] ./you-get -i http://www.bilibili.com/video/av2537307/index_11.html
Video Site: bilibili.com
Title:      【MJSK4字幕组】真假學園4 外伝1-12 蓝光版合集(11)
Type:       Flash video (video/x-flv)
Size:       762.03 MiB (799042740 Bytes)
```

The 11th partition is omitted.

```
λ  you-get-develop [develop] ./you-get -i http://www.bilibili.com/video/av2537307/index_12.html
Video Site: bilibili.com
Title:      【MJSK4字幕组】真假學園4 外伝1-12 蓝光版合集(12)
Type:       Flash video (video/x-flv)
Size:       753.12 MiB (789705477 Bytes)
```

The 12th partition is omitted.

The real size of the video should be 12P in 832.38 MiB (872816678 Bytes).

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/703)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -870,7 +870,7 @@ def script_main(script_name, download, download_playlist = None):
     -y | --extractor-proxy <HOST:PORT>       Use specific HTTP proxy for extracting stream data.
          --no-proxy                          Don't use any proxy. (ignore $http_proxy)
          --debug                             Show traceback on KeyboardInterrupt.
-         --json                              Output the information of videos in json text without downloading. 
+         --json                              Output the information of videos in json text without downloading.
     '''
 
     short_opts = 'Vhfiuc:nF:o:p:x:y:'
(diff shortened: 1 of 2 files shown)


soimort:
As I started to rethink, #377 is not really a good idea. For many videos on Bilibili, partitions are not necessarily bound together (might be different episodes, like [this one](http://www.bilibili.com/video/av2537307/)) it's problematic to merge all these partitions.

`you-get` should download the exact partition which user requires. For all other partitions, use the `-l` option to download the whole playlist (and no merging will be carried out).

PR #732, opened 2015-10-22
Remove support of SongTaste, Coursera, and a bunch more

soimort:
* **SongTaste** <http://www.songtaste.com>
    * No longer operating.
* **Coursera** <https://www.coursera.org>
    * I've followed a lot of courses on it, but never downloaded a single video from it. The extractor seems to be broken anyway.
* **Blip.tv** <http://blip.tv>
    * Redirected to <http://maker.tv>. I'm not interested in it.
* **Catfun** (喵星球) <http://www.catfun.tv>
* **VID48** <http://vid48.com>
* **VideoBam** <http://videobam.com>
    * Those are dead sites.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/732)

PR #710, opened 2015-10-18
[YouTube] Download DASH streams for VEVO videos

soimort:
Follow-up of #704 .
VEVO videos require encrypted signatures for DASH streaming. They may not be accessed via `dashmpd`.

Examples:
* https://www.youtube.com/watch?v=CWFUwdaT8kM
* https://www.youtube.com/watch?v=_HUVPglTfNY

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/710)

PR #705, opened 2015-10-17
[YouTube] Download subtitles / closed captions in SRT format

soimort:
Download YouTube videos together with subtitles / closed captions.

Since YouTube (also Google Video) uses its own non-standard XML format for storing timed text, and that is hardly ever supported by any media player, all downloaded captions will be converted to [SRT subtitle format](http://www.matroska.org/technical/specs/subtitles/srt.html) automatically.

(Fix #36, #235)

Example:
```
λ  you-get-develop [youtube-caption-tracks] you-get --itag=137 youtu.be/8r7SGQ4c4fQ
site:                YouTube
title:               【MV】Green Flash / AKB48[公式]
stream:
    - itag:          137
      container:     mp4
      quality:       1920 x 1080
      size:          86.6 MiB (90798818 bytes)
    # download-with: you-get --itag=137 [URL]

Downloading 【MV】Green Flash - AKB48[公式].mp4 ...
100.0% ( 86.6/86.6 MB) [========================================] 2/2
Merging video parts... Done.

Saving 【MV】Green Flash - AKB48[公式].ja.srt ...Done.
Saving 【MV】Green Flash - AKB48[公式].ko.srt ...Done.
Saving 【MV】Green Flash - AKB48[公式].zh-CN.srt ...Done.
Saving 【MV】Green Flash - AKB48[公式].en.srt ...Done.
```

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/705)

PR #704, opened 2015-10-17
[YouTube] Support DASH videos

soimort:
Support YouTube videos of 1080p or even higher qualities.
(Fix #395, #599)

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/704)

PR #734, opened 2015-10-23
Release 0.4.x

soimort:
Thanks to everyone who has ever contributed to `you-get`. Time for a new minor release now.
(the release was scheduled a week ago, with a few patches on interesting features I want, somehow it took a bit longer than expected)

It's a lot of commits (100+). If you want to get a quick overview of new features that have been added, take a look at #703, #704, #705, #706, #707, #708, #709, #710, #711, #712, #717, #719, #723, #730, #732. Also, grok the list of all relevant commits: https://github.com/soimort/you-get/compare/v0.3.36...0.4

There won't be a changelog to document and credit every improvement we made, because

giphy

## Notes for users

If you don't care about the actual code, just visit our website https://you-get.org/ and read the new user manual (will be online later today)

The issue page https://github.com/soimort/you-get/issues is only for reporting `you-get` relevant issues. It may not be used as a chatroom (and **don't post off-topic error reports under an irrelevant thread, these can be easily forgotten and overlooked by our developers; instead, open a standalone issue**). For general questions and discussions (you-get related or not related), feel free to use our IRC channel: https://gitter.im/soimort/you-get

Some issues may be fixed in _3 minutes_ by tweaking a regexp. Some others can take _[1048 days](https://github.com/soimort/you-get/issues/36)_, just as normal. Some will never get fixed, until the site is dead and gone. The best solution is always to fork the code and fix it yourself, and (if you like) send a PR.

## Notes for potential contributors

- General pull requests on fixing currently broken extractors are **more than welcome**.
- General pull requests on improving existing extractors are **welcome** (especially on switching them to `VideoExtractor` to enable format selection), but be careful not to break any URL pattern that is known to work.
- General pull requests on new sites are **still considered**, but no longer a priority.

If you have ever implemented an extractor for `you-get`, please take the responsibility to maintain it. Push access may be granted if you consistently send PRs to fix stuff (especially if you are able to maintain something I don't care about, but many users might really need, for example, iQIYI, Sohu, Sina)

Note: I'll be busy in the following months that I may not be able to respond to issues and review PRs in time. You may also @jackyzy823 for help, in case I'm away from keyboard.

## 0.4 roadmap

0.4 is a LTS release following 0.3.36. Aside from new features we implemented, it shares similar ideas with 0.3:

- Command-line interface (existing options) will be left unchanged. Internal implementations may change.
- New options and new site extractors may be added, but (hopefully) they won't break existing usage.
- We will still **not** introduce any external dependency other than FFmpeg. That is to say, during the lifespan of 0.4, we will stick with the Python standard library, and will not make a switch to any of these: Beautiful Soup, urllib3, or requests.

## On versioning

As you may noticed, the first release of 0.4.x is not `0.4.0`. A few explanations on these numbers:

- The major version is 0. (1 is reserved for the next generation of you-get, which may or may not happen)
- The minor version number increases whenever we make any breakthrough change.
- 111 is the commit number since the last minor version. I'll draft a new release on PyPI whenever the number seems cool to me, like 111, and possibly 128, 148 and 201. If you use `you-get` directly from the repo here, git pull should always get you to a newer version number (if there's any new commits)

That's all for now.

Have fun watching videos! :tv:

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/734)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,31 +1,82 @@
-/build/
-/dist/
-/MANIFEST
-*.egg-info/
+# Byte-compiled / optimized / DLL files
+__pycache__/
 *.py[cod]
+*$py.class
 
-_*/
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+*.egg-info/
+.installed.cfg
+*.egg
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*,cover
+.hypothesis/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
+*.log
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+target/
+
+# Misc
 _*
 *_
-
-*.bak
-*.download
-*.cmt.*
+README.html
+README.rst
 *.3gp
 *.asf
-*.flv
+*.download
 *.f4v
+*.flv
+*.gif
+*.jpg
 *.lrc
 *.mkv
 *.mp3
 *.mp4
 *.mpg
+*.png
+*.srt
 *.ts
 *.webm
-*.srt
-README.html
-README.rst
-
-*.DS_Store
-*.swp
-*~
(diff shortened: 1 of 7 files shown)


zhangn1985:
great job

cnbeining:
Great job.

Thank you so much for your brilliant work.

To make future works easier, I would suggest the format of ```VideoExtractor``` to be **set** before the release. Although hard, I am hoping the this would reduce(or even eliminate, which would be better) the chance of disruption between versions and forks.

As  ```common.py``` being updated as ```b277d8d``` , I am thinking it is time to reconsider #441 regarding of solving the problem of the need of rapid change and PRs coming to several websites(which I myself do not care about either) so individual developers can just maintain their taps on their own instead of raising PR every several days, also reduce the workload of maintainers without giving out too many push accesses in case of potential chaos or incorrect using of Git or Github. 

For the next few weeks my personal goal is trying to close as many issues as I can since most sites are quite stable, also try to catch up with commercial extractors.

May we make a difference.

PR #737, opened 2015-10-26
[dilidili] Add support

cnbeining:
```
 python3 you-get   --debug  http://www.dilidili.com/watch/9901/
Site:       dilidili
Title:      钢之炼金术师 FULLMETAL ALCHEMIST 第1话「钢之炼金术师」
Size:       180.35 MiB (189109341 Bytes)

Downloading 钢之炼金术师 FULLMETAL ALCHEMIST 第1话「钢之炼金术师」.mp4 ...
```

I've prepared ```stream_types``` for following developing.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/737)

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,7 @@ Use `--url`/`-u` to get a list of downloadable resource URLs extracted from the 
 | **Baidu<br/>百度贴吧** | <http://tieba.baidu.com/> |✓|✓| |
 | 爆米花网 | <http://www.baomihua.com/>     |✓| | |
 | **bilibili<br/>哔哩哔哩** | <http://www.bilibili.com/> |✓| | |
+| Dilidili | <http://www.dilidili.com/>    |✓| | |
 | 豆瓣     | <http://www.douban.com/>       | | |✓|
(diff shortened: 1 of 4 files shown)


soimort:
Just a heads up -

screenshot from 2015-10-27 02-25-40

Don't risk too much to have every existing video site covered.

cnbeining:
These days no one really cares what the heck does those 3rd ad parties would do.

image

At least with ```you-get``` we can somehow keep the end users from this kind of risk. In a way. More or less.

soimort:
The problematic site is embedded in a frame

```xml
<iframe id="iframe01" width=650 height=240 frameborder=0 src="http://www.xcmh.cc/out/out.php?key=钢之炼金术师FA"></iframe>
```

Of course, on `you-get` we don't need to worry about security issues this way (we're not trying to simulate a browser at this point). However, if a site appears to be phishing, it might not worth us the effort to implement anything for it. (still fine for this one, because the malware is on another domain xcmh.cc which is embedded in the page, not on dilidili.com itself)

PR #755, opened 2015-11-09
Windows build

soimort:
Starting from version **[0.4.136](https://github.com/soimort/you-get/releases/tag/v0.4.136)**, we will ship Windows binaries for every release of `you-get`, as I promised.

Those binaries were built with [PyInstaller](https://github.com/pyinstaller/pyinstaller), so no more Python installation is required. The download [you-get-0.4.136-win32.exe](https://github.com/soimort/you-get/releases/download/v0.4.136/you-get-0.4.136-win32.exe) is a standalone executable of `you-get`; the download [you-get-0.4.136-win32-full.7z](https://github.com/soimort/you-get/releases/download/v0.4.136/you-get-0.4.136-win32-full.7z) also contains ffmpeg and rtmpdump for convenience. If you are already using ffmpeg on Windows, you just need the `exe` file; otherwise, use the `7z` package.

Of course we don't have a GUI for `you-get`; instead, double click on the `you-get.bat` file, and it will open a command window where you'll able to run some commands like `you-get`, `ffmpeg` and `rtmpdump` there.
you-get

This (32-bit) executable has only been roughly tested on Windows 7. If you have issues running on other platforms, just let me know.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/755)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -555,7 +555,8 @@ class SimpleProgressBar:
 
     def update_received(self, n):
         self.received += n
-        bytes_ps = n / (time.time() - self.last_updated)
+        time_diff = time.time() - self.last_updated
+        bytes_ps = n / time_diff if time_diff else 0
         if bytes_ps >= 1048576:
             self.speed = '{:4.0f} MB/s'.format(bytes_ps / 1048576)
         elif bytes_ps >= 1024:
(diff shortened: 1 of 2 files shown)


guocdfeifei:
win10 not use

PR #773, opened 2015-11-26
腾讯extractor 也使用VideoExtractor

edward32tnt:
[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/773)

### DIFF
--- a/src/you_get/extractors/qq.py
+++ b/src/you_get/extractors/qq.py
@@ -1,27 +1,83 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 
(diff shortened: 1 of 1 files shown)


soimort:
Thanks for the pull request.

1. Could you please rewrite the commit message in English? Doing so may help a lot for anyone grepping the history of the project. A proper msg would be something like: `[qq] refactor extractor using VideoExtractor`
2. The new `qq.py` does not really provide anything interesting. It would be generally better to let it handle multiple formats / streams (because that's what a `VideoExtractor` is supposed to do).

PR #463, opened 2014-12-22
fix youku's bug #450 #390 translate from LiuLang  monkey-videos

linhua55:
翻译的 LiuLang 的monkey-videos

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/463)

PR #783, opened 2015-12-02
[Youku]Support full length of youku's video

### LINKED ISSUE #390: 无法下载完整程度的优酷视频
例子 http://v.youku.com/v_show/id_XNzM3OTM0OTEy.html
下载下来只有40分钟

linhua55:
Come from https://github.com/soimort/you-get/pull/463
fix #390 

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/783)

### DIFF
--- a/src/you_get/extractors/youku.py
+++ b/src/you_get/extractors/youku.py
@@ -7,6 +7,9 @@ from ..extractor import VideoExtractor
 import base64
 import time
 import traceback
+import urllib.parse
+import math
+import pdb
 
 class Youku(VideoExtractor):
     name = "优酷 (Youku)"
@@ -21,11 +24,32 @@ class Youku(VideoExtractor):
         {'id': 'mp4',    'container': 'mp4', 'video_profile': '高清'},
         {'id': 'flvhd',  'container': 'flv', 'video_profile': '标清'},
         {'id': 'flv',    'container': 'flv', 'video_profile': '标清'},
-        {'id': '3gphd',  'container': '3gp', 'video_profile': '标清（3GP）'},
+        {'id': '3gphd',  'container': 'mp4', 'video_profile': '标清（3GP）'},
     ]
 
-    def generate_ep(vid, ep):
-        f_code_1 = 'becaf9be'
+
+    def trans_e(a, c):
+        f = h = 0
+        b = list(range(256))
+        result = ''
+        while h < 256:
+            f = (f + b[h] + ord(a[h % len(a)])) % 256
+            b[h], b[f] = b[f], b[h]
+            h += 1
+        q = f = h = 0
+        while q < len(c):
+            h = (h + 1) % 256
+            f = (f + b[h]) % 256
+            b[h], b[f] = b[f], b[h]
+            if isinstance(c[q], int):
+                result += chr(c[q] ^ b[(b[h] + b[f]) % 256])
+            else:
+                result += chr(ord(c[q]) ^ b[(b[h] + b[f]) % 256])
+            q += 1
+
+        return result
+
+    def generate_ep(no,streamfileids,sid,token):
         f_code_2 = 'bf7e5f01'
 
         def trans_e(a, c):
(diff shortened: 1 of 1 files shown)


soimort:
The implementation looks good.

The copyright restricted videos (which used to be only accessible via the m3u8 method before), as discussed in [you-get#463](https://github.com/soimort/you-get/pull/463#issuecomment-68547897) and [youtube-dl#6377](https://github.com/rg3/youtube-dl/issues/6377), seem no longer to be an issue now. Here are some links to test: (previously not available)

- http://v.youku.com/v_show/id_XMTQ3MDcyNTYw.html
- http://v.youku.com/v_show/id_XNTc2NDk2MDMy

So, nice job.

However, the code is a bit tangled: `import urllib.parse` is not needed, and stuff for debugging purpose like `pdb` should be removed as you commit. The most important thing you forgot is, the code for processing passwords was commented out and never put back, so it will not work for password protected videos. (Test link: http://v.youku.com/v_show/id_XMzg1ODY1MTIw.html, password: `123`).

I'll do the clean-up later, so no further action is needed on your side.

linhua55:
Thanks.But it supports password protected videos.The code is at the handle part of `"error_code"`。That also should add "Referer" http header to http request.To fix #793.Maybe need some modification on `common.py`.

soimort:
`Referer: http://static.youku.com` worked fine for me just hours ago -- Now I'm getting 403 errors everywhere, as someone already mentioned in #793.

```
urllib.error.HTTPError: HTTP Error 403: Forbidden
```

Could this 403 be related to cookies? I suppose having only the referer is not enough.

linhua55:
Requesting the `play.youku.com` doesn't need cookies.But in original way,need to request `pl.youku.com`,And that needs cookies.

There are't `HTTP Error 403` in my test yet.  
What's the url you are testing?

soimort:
I can confirm that `k.youku.com` also requires cookies. And you will eventually get blocked.

screenshot from 2015-12-09 08-25-11

linhua55:
You can open the `k.youku.com` url in browser.
There are two ways to fix it.Depends on the `yxon=1` parameter in the end of `k.youku.com' url.
1.Now the `k.youku.com` url doesn't have `yxon=1` parameter,We can add http headers to download.But the `-p` argument in `you-get` may corrupt.
2.if we add `yxon=1` to the `k.youku.com`.The url will be '`k.youku.com/......&yxon=1`.And it will turn back a string contain a url of videos.You can test it in browser.

soimort:
> You can open the k.youku.com url in browser.

No, you can't, unless your browser already has its cookies (which is often the case if you have visited the Youku page in your browser before).

You should try with your browser's incognito mode, and see if things still work out fine even without any cookies.

linhua55:
Yes,the `k.youku.com` also need cookies.Use browser's incognito mode,it indeed give back `403 Error` to me.

src/you_get/extractors/youku.py
```diff
@@ -165,31 +217,45 @@ def extract(self, **kwargs):
...
+        streamfileid = self.streams_parameter[stream_id]['fileid']
+        for no in range(0,len(segs)):
+            k = segs[no]['key']
+            if k == -1:
+                log.e('Error')
+                exit()
+            fileId,ep  = self.__class__.generate_ep(no,streamfileid ,sid,token)
+#            pdb.set_trace()
+            m3u8  += 'http://k.youku.com/player/getFlvPath/sid/'+ sid
+            m3u8+='_00/st/'+ self.streams[stream_id]['container']
```
zhangn1985: ``_00`` i think it should be ``'_%02x' % no``

linhua55: Yes,you are right.

soimort: What difference does it make? It seems they are identical files.

zhangn1985: maybe it can help to resolve 403 issue, after ``Referer`` is added.

soimort:
:+1:

PR #784, opened 2015-12-02
[Yixia]Add Xiaokaxiu, redo Yixia logic

cnbeining:
These 2 sites are running by the same company, so I've redo their logics.

Not the same API though.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/784)


soimort:
Weird naming. :frowning:

```py
'weibo'      : 'miaopai',
[...]
'miaopai'    : 'yixia',
```

PR #794, opened 2015-12-08
youku download bug fix

Friday21:
add Referer when get 'http://play.youku.com/play/get.json?vid={vid}&ct=12' and record response cookie

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/794)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -251,6 +251,9 @@ def get_response(url, faker = False, youku_url=None):
         opener = getOpener(fake_headers)
         response = opener.open(url)
     else:
+        if cookies:
+            opener = request.build_opener(request.HTTPCookieProcessor(cookies))
+            request.install_opener(opener)
         response = request.urlopen(url)
 
     data = response.read()
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -79,6 +79,7 @@ SITES = {
 }
 
 import getopt
+import http
 import json
 import locale
 import os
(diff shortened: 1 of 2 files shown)


src/you_get/common.py
```diff
@@ -231,16 +232,28 @@ def undeflate(data):
...
+    return opener
 
+# DEPRECATED in favor of get_content()
+def get_response(url, faker = False, youku_url=None):
     if faker:
-        response = request.urlopen(request.Request(url, headers = fake_headers), None)
+        if youku_url:
+            fake_headers['Referer'] = youku_url
+        opener = getOpener(fake_headers)
+        response = opener.open(url)
```
zmwangx: You appear to be overriding all other installed openers here, including proxy settings. Don't do that! Adding a header is enough.

Friday21: @zmwangx  Here I am trying to store cookies , the old way did't seem work for me. I must get the response cookie when I am trying to download youku video,  can you show me how should I change it ?

zmwangx:
Sorry, I'm afraid I can't help here because I'm not personally familiar with `urllib` (I always let `requests` do the heavylifting for me), and I didn't have much time to mess around. There are two things I'm positively sure though:

1. As your PR currently stands, extractor proxy is ignored;
2. Your proposed workaround of adding a referer header and persisting the cookie does work.

Claim 1 was confirmed by looking at the response with a debugger. Claim 2 was confirmed by delegating to `curl` as a temporary workaround (yeah I was a bit pissed with all the complications here...) By the way, here's the `curl` proof-of-concept I'm referring to: https://gist.github.com/anonymous/d97766e46b40cd18cc77#file-common-py-L267-L285.

soimort:
As @zmwangx already pointed out, this PR is not very acceptable.

I would like to add that:
1. `get_response` has been marked as DEPRECATED in the code, so future PRs should really move onto improving / transferring to the recommended method (in this case, `get_content`). instead of trying to fix a deprecated thing.
2. For separation of concerns, do not introduce site-specific arguments like `youku_url` in `common.py` functions such as `get_html()`. If we have 10 more sites that we need to handle in the similar way as Youku, the argument list will soon be bloated.

Thanks for the PR, anyway.

PR #798, opened 2015-12-09
[youku]fix --player option

### LINKED ISSUE #796: [youku] --player option no longer works
If you play youku video using an external player such as mpc-hc. It no longer works.
Previously it worked like other sites.
Is this because youku is expecting "Referer: "Header or youku is expecting cookie?

linhua55:
fix #796

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/798)

### DIFF
--- a/src/you_get/extractors/youku.py
+++ b/src/you_get/extractors/youku.py
@@ -237,7 +237,7 @@ class Youku(VideoExtractor):
                 fileid    = fileid,
                 q         = q
             )
-            ksegs.append(u)
+            ksegs.append(json.loads(get_html(u+"&yxon=1"))[0]['server'])
 
         if not kwargs['info_only']:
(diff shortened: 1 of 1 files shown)


src/you_get/extractors/youku.py
```diff
@@ -237,7 +237,7 @@ def extract(self, **kwargs):
                 fileid    = fileid,
                 q         = q
             )
-            ksegs.append(u)
+            ksegs.append(json.loads(get_html(u+"&yxon=1"))[0]['server'])
```
soimort: 1. `get_html` is deprecated; we prefer to use `get_content` in future improvements.
2. Better style to encode the extra parameter `yxon` in the dict for `q`, instead of inserting as a string.

soimort:
I'll update this accordingly. Thanks for the fix!

PR #839, opened 2015-12-29
[youku] obtain cookies

### LINKED ISSUE #838: 无法下载 U酷 视频
$ you-get http://v.youku.com/v_show/id_XNDQzNTIyMDQ0.html?from=y1.6-100.3.1.a7c275ecd7b411e1a19e

you-get: [error] oops, something went wrong.
you-get: don't panic, c'est la vie. please try the following steps:
you-get:   (1) Rule out any network problem.
you-get:   (2) Make sure you-get is up-to-date.
you-get:   (3) Check if the issue is already known, on
you-get:         https://github.com/soimort/you-get/wiki/Known-Bugs
you-get:         https://github.com/soimort/you-get/issues
you-get:   (4) Run the command with '--debug' option,
you-get:       and report this issue with the full output.

是不是软件需要更新了？

linhua55:
fix #838
有bug ，请求视频地址不一定返回cookie。第一次请求时，一定返回。然后需要等待一段时间再次请求，才会返回。

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/839)

### DIFF
--- a/src/you_get/extractors/youku.py
+++ b/src/you_get/extractors/youku.py
@@ -141,6 +141,7 @@ class Youku(VideoExtractor):
                 self.download_playlist_by_url(self.url, **kwargs)
                 exit(0)
 
+        get_content(self.url)
         api_url = 'http://play.youku.com/play/get.json?vid=%s&ct=12' % self.vid
         api_url1 = 'http://play.youku.com/play/get.json?vid=%s&ct=10' % self.vid
         try:


soimort:
@linhua55 Why?

linhua55:
I think this is not a mature method.This will confuse some user when the request can't return cookies and throw out `HTTP 403 forbidden` error.And a cookie manager would be better.The feature of cookie manager would be:
  save the cookies for later use.
  renew the cookies when cookies expire.

soimort:
We can consider this as a temporary fix. If I understand correctly, it's Youku screwed up itself on cookies verification, not that it changed its encryption method or whatsoever.

linhua55:
Yes.This is a temporary fix.But it isn't perfect.

linhua55:
@soimort 
If we substitute `self.url` with `http://v.youku.com/v_show/`.The bug will disappear.
    get_content("http://v.youku.com/v_show/")

refer to  http://www.cnblogs.com/zhxilin/p/4993074.html#3337383

Issue #887, opened 2016-01-18
b站视频有些无法下载


kevinbakbak:
mulin ~ $ you-get --debug http://www.bilibili.com/video/av1727243/
you-get: version 0.4.245, a tiny downloader that scrapes the web.
you-get: ['http://www.bilibili.com/video/av1727243/']
Traceback (most recent call last):
  File "/usr/bin/you-get", line 9, in <module>
    load_entry_point('you-get==0.4.245', 'console_scripts', 'you-get')()
  File "/usr/lib/python3.5/site-packages/you_get/__main__.py", line 92, in main
    main(**kwargs)
  File "/usr/lib/python3.5/site-packages/you_get/common.py", line 1237, in main
    script_main('you-get', any_download, any_download_playlist, **kwargs)
  File "/usr/lib/python3.5/site-packages/you_get/common.py", line 1158, in script_main
    download_main(download, download_playlist, args, playlist, output_dir=output_dir, merge=merge, info_only=info_only, json_output=json_output, caption=caption)
  File "/usr/lib/python3.5/site-packages/you_get/common.py", line 1006, in download_main
    download(url, **kwargs)
  File "/usr/lib/python3.5/site-packages/you_get/common.py", line 1230, in any_download
    m.download(url, **kwargs)
  File "/usr/lib/python3.5/site-packages/you_get/extractors/bilibili.py", line 115, in bilibili_download
    title = unescape_html(title)
  File "/usr/lib/python3.5/html/__init__.py", line 130, in unescape
    if '&' not in s:
TypeError: argument of type 'NoneType' is not iterable

还有斗鱼TV不能调用vlc。加-d没任何反应...

zhangn1985:
斗鱼要回退2个提交
另外这个视频貌似要登陆

zhangn1985:
你也可以试试我的fork

cnbeining:
问题复现。

请带上你的cookie。

斗鱼问题请另开issue。

soimort:
> 请带上你的cookie。

kevinbakbak:
@cnbeining  @soimort    谢谢，暂时看战旗吧。望修复下斗鱼。。。
@zhangn1985  https://github.com/zhangn1985/you-get 试了，斗鱼一样不能看，还是谢谢帮助-.-

cnbeining:
斗鱼请另开issue。

2016年2月2日星期二，sowhatlph <notifications@github.com> 写道：

PR #894, opened 2016-01-22
zhanqi TV now can use old algorithm, so revert recent 2 comments

zhangn1985:
https://github.com/soimort/you-get/issues/883
https://github.com/soimort/you-get/issues/887


soimort:
Thanks! :+1:

PR #896, opened 2016-01-23
[qq] fix title for some videos

lilydjwg:
e.g. http://v.qq.com/cover/p/ps6mnfqyrfo7es3.html?vid=q0181hpdvo5
hope this won't break others, or we should try to recognize the URL
pattern (maybe apply to only URLs with "/cover/")?

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/896)


soimort:
Looks good to me. Thanks!

PR #898, opened 2016-01-23
Bar finetune

lilydjwg:
see the commits for more info.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/898)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -691,11 +691,13 @@ def download_urls(urls, title, ext, total_size, output_dir='.', refer=None, merg
     if len(urls) == 1:
         url = urls[0]
         print('Downloading %s ...' % tr(output_filename))
+        bar.update()
         url_save(url, output_filepath, bar, refer = refer, faker = faker, headers = headers)
         bar.done()
     else:
         parts = []
         print('Downloading %s.%s ...' % (tr(title), ext))
+        bar.update()
         for i, url in enumerate(urls):
             filename = '%s[%02d].%s' % (title, i, ext)
             filepath = os.path.join(output_dir, filename)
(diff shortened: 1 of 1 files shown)


soimort:
1. I'm not quite sure what this PR achieves. Can you please elaborate this a little bit by attaching a screenshot?
2. Your solution overflows the bar near the end of the downloading progress. (`bar_size` should be kept consistent)

screenshot from 2016-01-24 03-45-06

soimort:
Shrinking 42 to 38 doesn't look good, too.

**42** was chosen because it's the answer given by the following calculation:

```console
100.0% (999.9/999.9MB) ├████████████████████████████████████████┤[99/99] 1000 MB/s
|0                   24|                                        |0             18|
```

`24 + 18 = 42`.

Of course it won't fit videos larger than 1GB or that consist of more than 99 pieces, or even net speed faster than 9999MBps. But in practice, 42 is the best number we'd have. :satisfied:

lilydjwg:
The link in the first commit has over 300 pieces so it behaves like the buggy "eta" version of wget. It shouldn't overflow, or change during the process. I'll look into why shortly.

lilydjwg:
Here's the screenshot you asked for:


I've fixed the issue you've seen. I forgot about `self.bar_size` in other places.

soimort:
Well, I did not expect Letv could have a 3-digit number of pieces (#606 needs to be solved); in that case it all makes sense then.

soimort:
Thanks!

soimort:
I had to revert this (temporarily), as I'm still seeing the overflowing bar everywhere on large-sized videos (> 1GB). 

screenshot from 2016-02-04 19-39-39

Please free free to reopen.

soimort:
Will changing `38` simply to a bigger value work for you? (your solution works better for Letv videos with 100+ pieces, though)

lilydjwg:
I'll come up with a patch ensuring all variable-size parts don't get too
large later.
2016年2月5日 上午8:08，"Mort Yao" <notifications@github.com>写道：

PR #899, opened 2016-01-23
[ffmpeg] more check_call replacing wrong 'raise's

lilydjwg:
I also met this error message from ffmpeg:
```
[mp4 @ 0x562afd33f380] Malformed AAC bitstream detected: use the audio bitstream filter 'aac_adtstoasc' to fix it ('-bsf:a aac_adtstoasc' option with ffmpeg)
```

Adding the mentioned option fixes the conversion. But I don't know whether we should add this option into you-get.

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/899)

### DIFF
--- a/src/you_get/processor/ffmpeg.py
+++ b/src/you_get/processor/ffmpeg.py
@@ -170,11 +170,9 @@ def ffmpeg_concat_mp4_to_mp4(files, output='output.mp4'):
         params.append(output + '.txt')
         params += ['-c', 'copy', output]
 
-        if subprocess.call(params) == 0:
-            os.remove(output + '.txt')
-            return True
-        else:
-            raise
+        subprocess.check_call(params)
+        os.remove(output + '.txt')
+        return True
 
     for file in files:
         if os.path.isfile(file):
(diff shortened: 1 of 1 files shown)


soimort:
I prefer the let-it-crash approach when calling `ffmpeg`. If things go wrong (i.e. subprocess.call returns non-zero value), users will see the warning and have a chance to fix things on themselves. If you consider these calls to be "successful", all downloaded segments will be deleted as a consequence (which might not be a desired behavior).

lilydjwg:
No, `subprocess.check_call` just does what you intended: if the command isn't successful, it raises an `CalledProcessError` exception, with details about what the command was, and the exit code it returned.

The original handling is wrong because there is no exception to reraise, so Python raises a `RuntimeError`, with little information about what went wrong, so I have to change it.

zhangn1985:
@lilydjwg 
how about code below?
```

def launch_player(player, urls):
    import subprocess
    import shlex
    subprocess.call(shlex.split(player) + list(urls))
```

lilydjwg:
@zhangn1985 how's this code related to ffmpeg calls?

soimort:
Oh, I didn't realize it was `subprocess.check_call` instead of `subprocess.call` (I mistakenly thought you're just removing the indentation here). Iit looks fine.

soimort:
Thanks!

soimort:
@zhangn1985 here `launch_player` should not be using `subprocess.check_call` (we don't need to throw any exception when calling external players, I think)

PR #918, opened 2016-02-10
fix bar overflows

lilydjwg:
Fix the issue mentioned here: https://github.com/soimort/you-get/pull/898#issuecomment-180113961

I haven't test with real data yet, because currently I don't have the bandwidth to do so. Please give this a try :-)

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/918)


soimort:
Thanks! We'll see if it fits all types of videos soon.

soimort:
So far it looks good. Thanks!

PR #936, opened 2016-02-27
Octocat's tentacle stopped waving

soimort:
Our cute octocat on the website (https://you-get.org) used to wave nicely on the mouse `:hover` event (just like [this](http://tholman.com/github-corners/)), now it doesn't move anymore.

It seems CloudFlare is breaking the CSS code containing the `@` character (which is not an actual email address it intended to obfuscate). Since the site is hosted on GitHub pages, (currently) there's no direct way to use HTTPS with a custom domain other than employing a 3rd party CDN like CloudFlare or CloudFront; but that's not a "true HTTPS", as you see, the original HTML (I pushed on GitHub pages through `gh-pages` branch) has been altered mistakenly (by the CDN).

In case you're interested in using HTTPS with your own site on GitHub Pages, let's vote on this issue:
https://github.com/isaacs/github/issues/156

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/936)


cnbeining:
I think you can just...shut the obfuscation down in Cloudflare. This is optional and would not affect the CDN or anti-DDoS.

soimort:
Good idea to just shut it down (looking for that option now). `pandoc` has email obfuscation already; plus, we didn't leave any maintainer's email address on the site (or should we?)

cnbeining:
I don't think we have to put everyone's email there: should someone wants to reach us, they can open an issue; should they want to reach someone individually, we are always having our contacts on our Github user page...

And,

image

soimort:
Nailed it, thanks.

p.s. ping me if you'd finish the Chinese translation someday; I may port things from wiki to the main site there (it's more approachable than wiki)

PR #960, opened 2016-03-06
[.github] Add FAQ to issue

cnbeining:
Add FAQ to issue

[Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/960)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,7 +2,8 @@ Please make sure these boxes are checked before submitting your issue – thank 
 
 - [ ] You can actually watch the video in your browser or mobile application, but not download them with `you-get`.
 - [ ] Your `you-get` is up-to-date.
-- [ ] The issue is not yet reported on <https://github.com/soimort/you-get/issues> , <https://github.com/soimort/you-get/wiki/FAQ> or  <https://github.com/soimort/you-get/wiki/Known-Bugs>. If so, please add your comments under the existing issue.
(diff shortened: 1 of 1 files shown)


soimort:
Better to separate it out as an individual item, FAQ is not anything about issues, but common problems regarding general usage.

cnbeining:
Title misleading...

What I am trying to do is I want all users who want to raise any issue to **READ THE DAMNED FAQ** before asking any questions. I even drawn them a picture.

soimort:
I meant something like this:

- [x] Make sure you have read <https://github.com/soimort/you-get/wiki/FAQ> before reporting any issue.
- [ ] The issue is not yet reported on blabla ...

Sadly I doubt whether people are actually reading our ISSUE_TEMPLATE, leave alone all these extra links. (Well, this might the best we could do.)

Personally I consider "having to read the FAQ" as bad user experience. I must say It's not users' fault, but that Microsoft really needs to fix their **DAMNED `cmd.exe`**. Even I myself had a hard time doing testing and packaging for `you-get` under Windows.

cnbeining:
I really cannot stop myself from add up something like 

> failure to do so will result in public humiliation and disgrace

at the end of the issue template...

soimort:
> I really cannot stop myself from add up something like

You may not do so. Feel free to ignore any issues that are too "noob" for you to reply, and I'll handle the rest. (by copy and pasting the link to them)

cnbeining:
Well, no. Just saying.

cnbeining:
Close and make a real one.

PR #963, opened 2016-03-07
[VLook]Add support

cnbeining:
```
$ python3 you-get --debug 'http://www.vlook.cn/show/qs/YklkPTMyODEwMDE='
Site:       vlook
Title:      坏才刘科学:那些年不懂事，以为是青春痘就给抠掉了！---《坏才刘科学》
Size:       2.36 MiB (2470427 Bytes)

Downloading 坏才刘科学-那些年不懂事，以为是青春痘就给抠掉了！---《坏才刘科学》.mp4 ...
 0.0% (  0.0/  2.4MB) ├─────────────────────────────────────────┤[1/1] ^C
```

I do not really understand why people are paying for this extractor.

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -75,6 +75,7 @@ SITES = {
     'weibo'            : 'miaopai',
     'veoh'             : 'veoh',
     'vine'             : 'vine',
+    'vlook'            : 'vlook',
     'vk'               : 'vk',
     'xiami'            : 'xiami',
     'xiaokaxiu'        : 'yixia',
--- a/src/you_get/extractors/__init__.py
+++ b/src/you_get/extractors/__init__.py
@@ -63,6 +63,7 @@ from .veoh import *
 from .videomega import *
 from .vimeo import *
 from .vine import *
+from .vlook import *
 from .vk import *
 from .w56 import *
 from .xiami import *
--- a/src/you_get/extractors/vlook.py
+++ b/src/you_get/extractors/vlook.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+
+__all__ = ['vlook_download']
+
+from ..common import *
+import urllib.error
+
+def vlook_download_by_url(url, output_dir = '.', merge = False, info_only = False, **kwargs):
+    '''Source: Android mobile'''
+    if re.match(r'http://www.vlook.cn/show/qs/.+', url):
+        fake_headers_mobile = {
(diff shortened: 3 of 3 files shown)


soimort:
No offense, but this site seems totally cr*p for me.

screenshot from 2016-03-07 15-29-11

Does anyone really need to download videos from a site like this?

cnbeining:
Not all of them. 

I 've seen lots of requests at http://www.ckplayer.com/bbs/forum.php?mod=forumdisplay&fid=40 , so I would just put it here.

It's more or less like Miaopai or Xiaokaxiu, something like that.

soimort:
A quick question: do you need this yourself?

I've added a bunch of unpopular sites which I needed my own, that's fine because I use them on weekly/daily basis, so they are more likely to be properly maintained (by me). If you need to visit this site regularly, then it's fine, too, because it's guaranteed to be maintained (by you). However, if you don't even need this yourself, it's very questionable whether we should spend more time on this.

(I really need GitHub to have polls, so we can vote on a new site to implement; if many people request the same thing here, we may prioritize it rather than wasting time on something nobody needs)

For now, if anyone need this site, please reply this thread so I know it's worthwhile.

cnbeining:
Good question.

Let's add TAPs first so everyone can add their own extractors. This is also
useful for bleeding edge patchs that would require more testings or stuffs
changing extremely quick (like salt in iQiyi).

2016年3月7日星期一，Mort Yao <notifications@github.com> 写道：

soimort:
Taps are the best solution to this, I think.

I understand that different people have different needs, but we (as contributors) have very limited time to keep up with all these changing sites. So if we are to add something new, we must be able to devote some time to maintain them in the future. From my experience, if you try to implement an extractor for every site you randomly see on the Internet, that WILL exhaust your valuable time!

Oh, don't take this the wrong way, but _the more is not the better_.

cnbeining:
I will look into it myself, but I would appreciate help from your side
since you are more familiar with the structure of the code.

Many thanks.

2016年3月7日星期一，Mort Yao <notifications@github.com> 写道：

soimort:
> I will look into it myself, but I would appreciate help from your side
since you are more familiar with the structure of the code.

The current structure of codebase is a mess: mixture of OOP and non-OOP (legacy) code, inconsistent coding style, quick hacks and patches merged without much thought.
(Surely I'm the one responsible for this, but "just getting things done asap" was my philosophy for this project. As long as it can download something, it's good)

I won't have much time to clean up the code at this moment, but if you find an existing extractor is broken, feel free to fix it, I'd be thankful for that - that would be our No.1 priority!

PR #976, opened 2016-03-11
Add huaban.com support.

wenLiangcan:
Sample link: http://huaban.com/boards/24803867/

---
This change is [Review on Reviewable](https://reviewable.io/reviews/soimort/you-get/976)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,5 @@ _*
 *.ts
 *.webm
 *.xml
+/.env
+/.idea
--- a/README.md
+++ b/README.md
@@ -371,6 +371,7 @@ Use `--url`/`-u` to get a list of downloadable resource URLs extracted from the 
 | **Youku<br/>优酷** | <http://www.youku.com/> |✓| | |
 | 战旗TV   | <http://www.zhanqi.tv/lives>   |✓| | |
 | 央视网   | <http://www.cntv.cn/>          |✓| | |
+| 花瓣     | <http://huaban.com/>           | |✓| |
 
 For all other sites not on the list, the universal extractor will take care of finding and downloading interesting resources from the page.
 
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -24,6 +24,7 @@ SITES = {
     'fun'              : 'funshion',
     'google'           : 'google',
     'heavy-music'      : 'heavymusic',
+    'huaban'           : 'huaban',
     'iask'             : 'sina',
     'ifeng'            : 'ifeng',
     'imgur'            : 'imgur',
--- a/src/you_get/extractors/__init__.py
+++ b/src/you_get/extractors/__init__.py
@@ -21,6 +21,7 @@ from .freesound import *
 from .funshion import *
 from .google import *
 from .heavymusic import *
+from .huaban import *
 from .ifeng import *
 from .imgur import *
 from .infoq import *
--- a/src/you_get/extractors/huaban.py
+++ b/src/you_get/extractors/huaban.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python
+
+import json
+import os
+import re
+import traceback
+import urllib.parse as urlparse
+
+from ..common import *
+
+__all__ = ['huaban_download']
+
+site_info = '花瓣 (Huaban)'
+
+LIMIT = 100
+
+
+class EnhancedPiecesProgressBar(PiecesProgressBar):
+    BAR_LEN = 40
+
+    def update(self):
+        self.displayed = True
+        bar = '{0:>5}%[{1}] {2}/{3}'.format(
+            '', '=' * self.done_bar + '-' * self.todo_bar,
+            self.current_piece, self.total_pieces)
+        sys.stdout.write('\r' + bar)
+        sys.stdout.flush()
+
+    @property
+    def done_bar(self):
+        return self.BAR_LEN // self.total_pieces * self.current_piece
+
+    @property
+    def todo_bar(self):
+        return self.BAR_LEN - self.done_bar
+
+
+class Board:
+    def __init__(self, title, pins):
+        self.title = title
+        self.pins = pins
+        self.pin_count = len(pins)
+
+
+class Pin:
+    host = 'http://img.hb.aicdn.com/'
+
+    def __init__(self, pin_json):
+        img_file = pin_json['file']
+        self.key = img_file['key']
+        self.url = urlparse.urljoin(self.host, self.key)
+        self.ext = img_file['type'].split('/')[-1]
+
+
+def construct_url(url, **params):
+    param_str = urlparse.urlencode(params)
+    return url + '?' + param_str
+
+
+def extract_json_data(url, **params):
+    url = construct_url(url, **params)
+    html = get_content(url, headers=fake_headers)
+    json_string = match1(html, r'app.page\["board"\] = (.*?});')
+    json_data = json.loads(json_string)
+    return json_data
+
+
+def extract_board_data(url):
+    json_data = extract_json_data(url, limit=LIMIT)
+    pin_list = json_data['pins']
+    title = json_data['title']
+    pin_count = json_data['pin_count']
+    pin_count -= len(pin_list)
+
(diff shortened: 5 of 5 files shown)


soimort:
Firstly, thanks for the pull request.

A rough review:

Your implementation of `huaban_download_board()` perfectly ignored the current structure of `you-get` - it has its own version of progress bar, own file handling, own `stdout` printing/logging, and even own `KeyboardInterrupt` exception catching (why?).

The code is not wrong - but as it differs quite a lot from `you-get`'s coding style, it will make itself extremely hard to maintain. If we improve any high-level routine in `common,py`, the code here will not benefit from our changes much, or even risk to break (e.g. it relies on some low-level API like `url_save`, which is not supposed to be used by extractors directly).

Especially, if you want to propose an enhancement to the current progress bar, or anything else that could be shared among all extractors, please consider doing a patch on `common.py`, instead of putting them all together specific to one site, isolated.

wenLiangcan:
There are two reasons make me feel hard to fit the download process of `Huaban` into the workflow provided by `common.py`.

First, there maybe multiple types of images in one "board". `download_urls` accepts only one extension name for a whole stream, and also `print_info`.

The second one is new images will be added into a "board". I want to continuously download images from a "board", but the default naming scheme is based on current number of files to be downloaded.

The reasons above explained why I have my own versions of file handling and logging.

For the progress bar, I wish its animation is based on the count of pieces of files rather than the size, because it isn't worth to make hundreds of connections to check the images' size of a "board". `SimpleProgressBar`'s animation is based on size and `PiecesProgressBar` just count down the number.

Lastly, it seems I misunderstood the exception handling strategy in [youku.py](https://github.com/soimort/you-get/blob/develop/src/you_get/extractors/youku.py#L121), it raise the `KeyboardInterrupt` but I bypass it. :disappointed:

soimort:
> For the progress bar, I wish its animation is based on the count of pieces of files rather than the size, 
> because it isn't worth to make hundreds of connections to check the images' size of a "board".

That is a very good point, and that's why I prefer you to make those changes in `common.py`, rather than do it in your own `huaban.py`.

`download_urls` does support only one extension, and you're also right, that the default naming scheme is based on the number of pieces (starting from 0). I haven't tested your PR on a real machine yet, but I guess you could just call `download_urls` as many times as want; doing that won't look very nice, though.

Or, if you can abstract some of these functionalities as common routines, put them into a common place (`common.py`) so later implemented extractors can benefit from them.

> Lastly, it seems I misunderstood the exception handling strategy in youku.py

Oh, forgot to say, but the `KeyboardInterrupt` in `youku.py` is used to skip a certain video when downloading a playlist (so if you have a video you don't like, `Ctrl-C` will let you skip it, instead of terminating the whole downloading process)

wenLiangcan:
> you could just call download_urls as many times as want

Yes, that will makes things more simple and I no longer need to care about the progress bar, but as you said, it is not so good looking.

This is the output based on my customized version file handling code:
```
you-get http://huaban.com/boards/24803867/
Site:       花瓣 (Huaban)
Title:      爱上咖啡的味道

Downloading 105 images in 爱上咖啡的味道 ...
     %[====------------------------------------] 8/105
```

and this is the output of "calling `download_urls` for one image at a time":
```
 you-get --debug http://huaban.com/boards/24803867/
Downloading 爱上咖啡的味道 (603832678).jpeg ...
 0.0% (  0.0/  infMB) ├────────────────────────────────────────────────┤[1/1]  208 kB/s

Downloading 爱上咖啡的味道 (579557455).jpeg ...
 0.0% (  0.1/  infMB) ├────────────────────────────────────────────────┤[1/1]  557 kB/s

Downloading 爱上咖啡的味道 (579557420).jpeg ...
 0.0% (  0.1/  infMB) ├────────────────────────────────────────────────┤[1/1]  514 kB/s
```
Do you really prefer this way? If so, I push the new commit.

soimort:
Actually I prefer the progress bar (probably other things also) to be fixed in `common.py`. For now, removing `EnhancedPiecesProgressBar`and simply making use of `download_urls()` will do the job.

I don't have time to refactor this part of `common.py`, so let's keep things simple that if I want to rewrite the whole codebase someway, simpler extractors would be easier to improve, i.e., the shorter, the better.

(Tip: it's completely up to you whether you would intend to change that, since you can always keep it as your own fork of `you-get`.)

Making hundreds of HEAD requests to check for all pieces' sizes, as you said, is inefficient for image downloading. The cause is, `you-get` was designed to be a video downloader (normally there won't be too many pieces for a video), so handling images wasn't my concern at the very beginning. Certainly, we have extended its scope to image downloading now.

wenLiangcan:
Yes, you are right, keeping simple is the best choice.

soimort:
Looks good. Could you please rebase this branch?

wenLiangcan:
I know this is a stupid question, but, how could I do that?

I read the README of this project and done the following steps before making the last commit:
```
git remote add upstream https://github.com/soimort/you-get.git
git fetch upstream
git rebase upstream/develop
```

I have a few experiments of opening PRs for other projects. After my code being merged, I would use the commands above to update my forked repo from the upstream, but I never do a rebase before the merging, so I'm not sure what that means : \

soimort:
Normally `rebase` is used *before* merging, since it rewrites all your commits you made ahead of the base branch, in this case, `soimort/develop`.

I can still merge your commits without rebasing; but it'd be better to make sure we have a tidy "linear" history, and that your PR will end up with a nicer look (purple "Merged" sign instead of a red "Closed" sign)

Please post your `git branch -va` before I can tell what the problem is.

wenLiangcan:
Here are the outputs:

```
git branch -va
* develop                   2a1a541 [Huaban] Refactoring.
  remotes/origin/0.4        fb8acf4 version 0.4.111
  remotes/origin/HEAD       -> origin/develop
  remotes/origin/develop    2a1a541 [Huaban] Refactoring.
  remotes/origin/gh-pages   741a354 version 0.4.324
  remotes/origin/master     bf9910a version 0.4.324
  remotes/upstream/0.4      fb8acf4 version 0.4.111
  remotes/upstream/develop  2a80fb5 Merge branch 'zhangn1985-for-up' into develop (fix #906)
  remotes/upstream/gh-pages 2062c68 version 0.4.330
  remotes/upstream/master   bf495b1 version 0.4.330
```

soimort:
This one is out-of-date with the base branch:

```
  remotes/upstream/develop  2a80fb5 Merge branch 'zhangn1985-for-up' into develop (fix #906)
```

wenLiangcan:
There's no conflicts after rebasing, did I rebased correctly?:

```
git fetch upstream
git rebase upstream/develop
```

I need to force push my commits otherwise I would get the following errors:
```
To git@github.com:wenLiangcan/you-get.git
 ! [rejected]        develop -> develop (non-fast-forward)
error: failed to push some refs to 'git@github.com:wenLiangcan/you-get.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

soimort:
GitHub still shows me "This branch is out-of-date with the base branch", but it's good to merge now. Thanks!

(it's ok to force push to a temporarily branch, but you might want to use something other than `develop` next time)

wenLiangcan:
> GitHub still shows me "This branch is out-of-date with the base branch"

I found two commits between mine and your merging commit, seems they are added after my rebasing, maybe that is the reason?

And thanks for the suggestion and your patience.

soimort:
> I found two commits between mine and your merging commit, seems they are added after my rebasing, maybe that is the reason?

Since there's no more conflict, it should not stand out-of-date anymore. If that reason is true, then whenever someone pushes some new commits to the upstream, every existing unmerged Pull Request should've been rendered outdated. That's obviously not the case. I don't know what GitHub's mechanism is; maybe they just don't bother rechecking a PR again even if it's updated.

> And thanks for the suggestion and your patience.

And you're welcome.:+1:

PR #999, opened 2016-03-20
Enhanced debug logging: Print URL of get_content call when --debug is enabled

zmwangx:
I know this PR is unsolicited and may be controversial (especially in terms of implementation), but the undeniable fact is currently the `--debug` option doesn't live up to its name. It only enables a traceback upon failure, which is useless most of the time (at least for me) without dropping into a debugger.

Most errors I have ever encountered occur at the extraction stage, so it's essential to know the URLs that you-get is trying to retrieve and parse in order to determine why it fails to do so. It also helps to know what's being retrieved in the case of zero failure yet seemingly hanging forever (happens to me often enough — connection to Mainland is really bad). AFAIK `get_content` is the most common and non-deprecated function used in the extraction stage, so in this PR I'm printing the URL of each `get_content` call when `--debug` is enabled. Here's an example:

```
> you-get -u -d http://v.youku.com/v_show/id_XMTUwNTkyOTQwNA\=\=.html
[DEBUG] get_content: http://play.youku.com/play/get.json?vid=XMTUwNTkyOTQwNA==&ct=10
[DEBUG] get_content: http://play.youku.com/play/get.json?vid=XMTUwNTkyOTQwNA==&ct=12
[DEBUG] get_content: http://k.youku.com/player/getFlvPath/sid/4458453567965121132a9_00/st/flv/fileid/030002010056EE397AB0EC11E60AB5BF5D6B79-F671-7D3E-FF81-C960F44E973B?ev=1&ep=diaREk2IVcsG4irZjT8bZi3hdiYGXP4J9h%2BFg9JjALshTp3KmjfWtu%2BwO4xDF4wfclN0FZmHrqThGEcdYYZFrmkQ2DqrTPqWiPji5d5awpB2Yh1De8rUxlScQjL4&K=3eedbd22d6eabeeb261ee89d&yxon=1&oip=1657983508&ctype=12&token=2129
site:                优酷 (Youku)
title:               Test
stream:
    - format:        flvhd
      container:     flv
      video-profile: 标清
      size:          0.0 MiB (28622 bytes)
    # download-with: you-get --format=flvhd [URL]

Real URLs:
http://103.38.56.154/youku/6573FE04A9D3E7A94929F6296/030002010056EE397AB0EC11E60AB5BF5D6B79-F671-7D3E-FF81-C960F44E973B.flv
```

Of course this debug logging mechanism may be extended to other functions too. The output format could be tweaked. (Note that I'm using `logging.debug` instead of `util.log.d` because the latter is unconditional, and passing an extra parameter around is too much trouble.)

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/999)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -91,6 +91,7 @@ SITES = {
 import getopt
 import json
 import locale
+import logging
 import os
 import platform
 import re
@@ -297,6 +298,8 @@ def get_content(url, headers={}, decoded=True):
     Returns:
         The content as a string.
     """
+
+    logging.debug('get_content: %s' % url)
 
     req = request.Request(url, headers=headers)
     if cookies:
@@ -1032,6 +1035,8 @@ def script_main(script_name, download, download_playlist, **kwargs):
               % get_version(kwargs['repo_path']
             if 'repo_path' in kwargs else __version__))
 
+    logging.basicConfig(format='[%(levelname)s] %(message)s')
+
     help = 'Usage: %s [OPTION]... [URL]...\n\n' % script_name
     help += '''Startup options:
     -V | --version                      Print version and exit.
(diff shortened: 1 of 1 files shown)


soimort:
Honestly speaking, I've not needed debug messages like this; leaving `tcpdump` open and you'll get a bit of everything underhood (at least for HTTP-only sites like Youku and Tudou, which is commonly the case for most Chinese websites).

However, it could be easier to tell in which URL you're stuck purely from the debug message of `you-get` itself. And it works for HTTPS links.

(the logging mechanism in `you-get` was lame; so it's ok to call `logging.debug` directly. This might change in the future)

Thanks for the contribution. :+1:

PR #1011, opened 2016-03-25
fix to handle new domain of douyu.com (was douyutv.com)

botusun:
fix to handle new domain of douyu.com (was douyutv.com)

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1011)


soimort:
I'm waiting for someone to confirm this. (or maybe we should keep both?)

```py
    'douyutv'          : 'douyutv',
    'douyu'            : 'douyutv',
```

zhangn1985:
they are same.

soimort:
Thanks!

PR #1028, opened 2016-04-04
Simplify logic via ternary if

cclauss:
Simplify logic via ternary if...  The () were required!!  Now passes tests.

---

This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1028)

### DIFF
--- a/src/you_get/__main__.py
+++ b/src/you_get/__main__.py
@@ -54,11 +54,8 @@ def main_dev(**kwargs):
                 # Display version.
                 log.println("you-get:", log.BOLD)
                 log.println("    version:  {}".format(__version__))
-                if head is not None:
-                    log.println("    branch:   {}\n    commit:   {}".format(*head))
-                else:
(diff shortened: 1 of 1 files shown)


soimort:
1. Could you please remove all the instruction message, and add a short explanation why your patch should take place, instead?
2. From Travis, I can see that this change will break Python 3.2, 3.3 and 3.4 compatibility, syntactically.

soimort:
Well, from what I (shallowly) understand, this patch shall make use of the [PEP 448](https://www.python.org/dev/peps/pep-0448/) implementation in **Python 3.5**. It is purely a syntactic addition, and it breaks compatibility with all previous Python versions. `you-get` is yet meant to be working across all major Python 3 versions, so we can't sacrifice all those *just for* a syntax sugar.

cnbeining:
We are seeing lots of aggressive changes in Python 3.4 and 3.5. Let's wait until Python 3.4 dies...

cclauss:
The () were required!! Now passes tests.

PR #1032, opened 2016-04-05
tightening up code

cclauss:
Remove 50+ lines of code while providing the same functionality with more performance and an easier read.

Avoid using `{}` as a default parameter.  See: http://docs.python-guide.org/en/latest/writing/gotchas/

---

This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1032)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -267,7 +267,7 @@ def get_location(url):
     # not to do that
     return response.geturl()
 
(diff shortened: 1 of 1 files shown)


soimort:
Maybe I'd rather avoid excessive code reformatting like this. It looks slightly better though, it will scrape the history and make `git blame` our code bitter...

But I'll take a look. Thanks.

Issue #499, opened 2015-03-02
Letv Failed


mysucklife:
Letv just downloading the ads, not the video, please fix it ASAP, thank you!

jackyzy823:
pls update to the latest version of you-get

for me ,it works.
see below:
$ python3 you-get http://www.letv.com/ptv/vplay/835594.html
Current Video Supports:
	--format 1300 <URL>
	--format 1000 <URL>
	--format 350 <URL>
Video Site: LeTV.com
Title:      倚天屠龙记邓超版02
Type:       MPEG-4 video (video/mp4)
Size:       321.67 MiB (337292868 Bytes)

Downloading 倚天屠龙记邓超版02.mp4 ...
  3.6% ( 11.7/321.7MB) [==                                      ] 10/160

SuperN0va:
下载没问题，但是合并有问题

allencch:
Merge problem, I solved with "cat" (concatenate) into one file. But this will lose the video duration information. So, I run the following command,
`ffmpeg -i combined.mp4 -c copy -bsf:a aac_adtstoasc fixed.mp4`

soimort:
Closing.

Please open another issue if it's still relevant, remember to include the video URL when reporting (otherwise there's no way we can fix it!)

PR #1041, opened 2016-04-07
Add -bsf:a aac_adtstoasc when merging

cnbeining:
As in #813 , #499  , #899 .

The question is: is this **harmless**?

@nareix

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1041)

### DIFF
--- a/src/you_get/processor/ffmpeg.py
+++ b/src/you_get/processor/ffmpeg.py
@@ -169,7 +169,7 @@ def ffmpeg_concat_mp4_to_mp4(files, output='output.mp4'):
 
         params = [FFMPEG] + LOGLEVEL + ['-f', 'concat', '-safe', '-1', '-y', '-i']
         params.append(output + '.txt')
-        params += ['-c', 'copy', output]
+        params += ['-c', 'copy', '-bsf:a', 'aac_adtstoasc', output]
 
         subprocess.check_call(params)
         os.remove(output + '.txt')


yan12125:
> is this **harmless**?

ffmpeg spams if the source stream is not in adts format. An example is rg3/youtube-dl#5591.

nareix:
tudou is ok after modified (not-adts aac flv -> mp4)

2016-04-08 12:19 GMT+08:00 Yen Chi Hsuan <notifications@github.com>:

yan12125:
I was wrong. ffmpeg warns for non-AAC streams. For non-ADTS AAC streams, it just does nothing. The example above has an MP3 audio stream.

soimort:
> is this harmless?

I'm not sure either. `ffmpeg_concat_mp4_to_mp4()` could be extended so that each extractor (e.g. `tudou.py`) can pass its own arguments for merging. Some sites were known to send malformed streams, so the options might be better to be handled on an individual basis (to prevent unintended effects on others).

A side note: don't change permissions for the file (644 -> 755).

cnbeining:
With ```youtube-dl```, they just put this on by default:

https://github.com/rg3/youtube-dl/blob/master/youtube_dl/postprocessor/ffmpeg.py#L270-L273

soimort:
Just restore the file permission to default 644.

cnbeining:
I will probably redo this PR later.

redstoneleo:
根据我的测试，还有这位的确认https://github.com/soimort/you-get/issues/813#issuecomment-293594575，高版本FFmpeg的不需要这个参数也能行

PR #1051, opened 2016-04-11
Extend unit tests for extractors

azizalghamdi:
Extended unit tests for youtube, xiami, vine, instagram, twitter, soundcloud, pinterest, tumbler, and ted extractor modules (26 test cases).

Each of the above websites have their test cases under one class (encapsulation).

Quick branch coverage was used to highlight the need for more unit test cases (using coverage.py).
Following is a table showing the improvements in coverage due to the extended unit tests:

| File | Before  | After |
| --- | --- | --- |
| src/you_get/extractors/youtube.py | 42% | 57% |
| src/you_get/extractors/xiami.py | 8% | 54%|
| src/you_get/extractors/vine.py | 30% | 80%|
| src/you_get/extractors/instagram.py |25% |92%|
| src/you_get/extractors/twitter.py |9% |64%|
| src/you_get/extractors/soundcloud.py |32%| 91%|
| src/you_get/extractors/pinterest.py|43%| 100%|
| src/you_get/extractors/ted.py|28%|84%|
| src/you_get/extractors/tumblr.py |10%| 39%|

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -20,11 +20,3 @@ class YouGetTests(unittest.TestCase):
 
     def test_mixcloud(self):
         mixcloud.download("http://www.mixcloud.com/DJVadim/north-america-are-you-ready/", info_only=True)
-
-    def test_vimeo(self):
-        vimeo.download("http://vimeo.com/56810854", info_only=True)
-
-    def test_youtube(self):
-        youtube.download("http://www.youtube.com/watch?v=pzKerr0JIPA", info_only=True)
-        youtube.download("http://youtu.be/pzKerr0JIPA", info_only=True)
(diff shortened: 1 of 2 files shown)


soimort:
Too many test cases will slow down our code integration with Travis and increase the probability of temporary failure (due to inaccessible servers or network status), that's why I haven't added more tests throughout years, in spite of our growing number of extractors.

Also, from my experience, extractors break down from time to time. If we can't fix any broken extractor immediately, then the whole status will be shown as "Failed", which would render our Travis tests for PR/commits useless, until we fix that specific site. Since Travis CI is mainly used for guaranteeing the PRs actually run (across various Python 3.x versions), it's actually wiser not to do intensive tests on extractors. A small number for commonly-used sites (e.g. YouTube) would suffice.

PR #1086, opened 2016-04-27
Update file you-get change env from `python` to `python3`

liudanking:
Fix issue `can't open file 'you-get': [Errno 2] No such file or directory` on OSX when run you-get with command `python3 you-get -h`

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1086)


soimort:
:+1:

Issue #1098, opened 2016-05-03
搜狐视频无法播放


zjb861107:
搜狐视频使用you-get无法播放，浏览器中是可以的。地址：http://tv.sohu.com/20141105/n405801469.shtml
命令行及输出：
you-get -p mpv --debug http://tv.sohu.com/20141105/n405801469.shtml
you-get: version 0.4.390, a tiny downloader that scrapes the web.
you-get: ['http://tv.sohu.com/20141105/n405801469.shtml']
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/bin/you-get", line 11, in <module>
    sys.exit(main())
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/you_get/__main__.py", line 92, in main
    main(**kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/you_get/common.py", line 1274, in main
    script_main('you-get', any_download, any_download_playlist, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/you_get/common.py", line 1187, in script_main
    download_main(download, download_playlist, args, playlist, output_dir=output_dir, merge=merge, info_only=info_only, json_output=json_output, caption=caption)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/you_get/common.py", line 1030, in download_main
    download(url, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/you_get/common.py", line 1267, in any_download
    m.download(url, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/you_get/extractors/sohu.py", line 41, in sohu_download
    host = info['allot']
KeyError: 'allot'

zhangn1985:
@soimort-bot please test: http://tv.sohu.com/20141105/n405801469.shtml

zhangn1985:
这个是因为最高清的视频信息是无效的，跳过就可以了。

PR #1099, opened 2016-05-03
ignore invalid video format

zhangn1985:
fix: #1098

in some case, highest video format is not valid, so just skip it, and go through to next available.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1099)


zhangn1985:
@soimort I just find a mistake in my 1st pr, so I use git push -f to force update.

if it's not ok, i will re-submit my patch.

soimort:
Force push is perfectly fine as long as you're working on a branch of your own.

:+1:

PR #1128, opened 2016-05-13
Print merged filename after successful merge

zmwangx:
After a successful merge, print `Merged into <filename>` instead of just `Done.`.

The rationale for this is that the `Downloading <filename> ...` message could be confusing because the extension given there is usually not the extension in the final merged file (see `get_output_filename` for extension mapping), so it's worthwhile to communicate to the user what the actual output file is.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1128)


soimort:
Makes sense :+1:

PR #1140, opened 2016-05-17
[MGTV]Fix some video cannot be downloaded in full length

cnbeining:
This is intended to fix links like http://www.mgtv.com/v/3/168294/f/3167498.html .

Anyone in Mainland can give me more test results? Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1140)

### DIFF
--- a/src/you_get/extractors/mgtv.py
+++ b/src/you_get/extractors/mgtv.py
@@ -7,6 +7,7 @@ from ..extractor import VideoExtractor
 from json import loads
 from urllib.parse import urlsplit
 from os.path import dirname
+import re
 
 class MGTV(VideoExtractor):
     name = "芒果 (MGTV)"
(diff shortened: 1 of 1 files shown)


soimort:
Inferred from the parameter, with `arange=` you get partial ranges; without it you get the whole thing. Apparent naming.

(Just saying; didn't test. Merge this if it works for you.)

cnbeining:
For some links (like the one I provided) MGTV would send this tiny little thing at the end of the URL to make it like a preview.

And people are paying for this little modification....sigh

It is really a pain to use Unblock Youku, Shadowsocks and Hola together. I will leave it here for those in Mainland to give a shot.

soimort:
> For some links (like the one I provided) MGTV would send this tiny little thing at the end of the URL to make it like a preview.

LOL.

I wish Youku was implemented that way.

cnbeining:
I am having successful report from Mainland China.

So I think it should be OK to merge this PR....

PR #1169, opened 2016-05-26
fix path error

4ft35t:
```bash
ln -s /path/to/you-get/you-get /usr/local/bin/you-get
```
execute you-get will get error
```python
➜  ~ you-get
Traceback (most recent call last):
  File "/usr/local/bin/you-get", line 9, in <module>
    import you_get
ImportError: No module named 'you_get'
```
use os.path.dirname(os.path.realpath(\_\_file\_\_)) get the soft link's real path to fix this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1169)

### DIFF
--- a/you-get
+++ b/you-get
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import os, sys
 
-_srcdir = 'src/'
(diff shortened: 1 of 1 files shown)


cnbeining:
GJ.

PR #1192, opened 2016-06-05
[Bilibili] download specified P according to the link

smart-techs:
I think the current behavior does not make sense. Sometimes I only want to download or play the specified P ,such as movie collection.

Can we do that according the link?
like that:
for http://www.bilibili.com/video/av3278560/index_1.html  , only download the first
for http://www.bilibili.com/video/av3278560/ ,download all parts
i'm a newcomer to python :)

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1192)

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -11,11 +11,13 @@ from .youku import youku_download_by_vid
 import hashlib
 import re
 
-appkey='f3bb208b3d081dc8'
+appkey = 'f3bb208b3d081dc8'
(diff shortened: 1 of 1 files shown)


soimort:
Looks good :+1: 

Would you mind amending the commit message to make it a more meaningful one?

smart-techs:
@soimort Of course not!

smart-techs:
@soimort Done.

soimort:
Thanks for the contribution!

PR #1231, opened 2016-06-28
enable socks5 proxy like shadowsocks

hjk41:
This is to enable socks proxy like shadowsocks. I added options -s and --socks-proxy to specify socks proxy. The patch depends on PySocks (imported only when --socks-proxy is specified). Tested with Youtube and works fine.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1231)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -1071,12 +1071,13 @@ def script_main(script_name, download, download_playlist, **kwargs):
     -x | --http-proxy <HOST:PORT>       Use an HTTP proxy for downloading.
     -y | --extractor-proxy <HOST:PORT>  Use an HTTP proxy for extracting only.
          --no-proxy                     Never use a proxy.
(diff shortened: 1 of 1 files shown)


soimort:
An adaptation of https://github.com/rg3/youtube-dl/pull/9287 should be preferable, but depending on PySocks looks fine to me in this case. (for its usefulness in resolving #1046 )

Thanks!

Nicksxs:
it still cannot work 
image
win10 x64
pip 7.1.2 from d:\python35\lib\site-packages (python 3.5)
Python 3.5.1

zhangn1985:
似乎 用pyinstaller打包好的Py程序，需要把依赖一并打包。

Nicksxs:
@zhangn1985 我打包develop分支的代码手动安装的，依赖打包要怎么做

zhangn1985:
先安装依赖，再打包就好了。

PR #1268, opened 2016-07-10
[SHOWROOM] Do polling on offline broadcasts

soimort:
When a show has not started yet, `you-get` summarily fails with an error message:

```console
$ yg -di https://www.showroom-live.com/48_Rena_Kato
[DEBUG] get_content: https://www.showroom-live.com/48_Rena_Kato
[DEBUG] get_content: https://www.showroom-live.com/api/live/streaming_url?room_id=61567&_=1468130011718
you-get: Cannot find any live URL! Maybe the live have ended or haven't start yet?
```

So it only works when the live show has _already started_, but that would be usually too late. Better to do some polling on that. (so `you-get` can record the stream as soon as it is available)

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1268)

### DIFF
--- a/src/you_get/extractors/showroom.py
+++ b/src/you_get/extractors/showroom.py
@@ -5,7 +5,7 @@ __all__ = ['showroom_download']
 from ..common import *
 import urllib.error
 from json import loads
-from time import time
+from time import time, sleep
 
 #----------------------------------------------------------------------
 def showroom_get_roomid_by_room_url_key(room_url_key):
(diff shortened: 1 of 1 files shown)


cnbeining:
image

soimort:
@cnbeining Comments, please.
Are you ok with this or do you prefer a let-it-crash pattern?

cnbeining:
I am OK with both. Yes it do make sense that we don't actually know when
the live shall start.

2016年7月10日星期日，Mort Yao <notifications@github.com> 写道：

PR #1278, opened 2016-07-13
	modified:   src/you_get/extractors/__init__.py

natugrey:
add freebuf package
modified:   src/you_get/extractors/__init__.py
new file:   src/you_get/extractors/freebuf.py
modified:   tests/test.py

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1278)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -21,6 +21,7 @@ SITES = {
     'facebook'         : 'facebook',
     'fc2'              : 'fc2video',
     'flickr'           : 'flickr',
+    'freebuf'          : 'freebuf',
     'freesound'        : 'freesound',
     'fun'              : 'funshion',
     'google'           : 'google',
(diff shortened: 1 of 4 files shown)


cnbeining:
Thank you very much. Great start for joining this community of ```you-get```!

Some very quick thoughts:

- Make a more meaningful name and write more meaningful messages.

- Do not modify existing test cases. They are necessary for CI.

- You do not need to add a test for this: we do not want a FULL coverage just in case some minor problem would block the developing.

- It would be more helpful if some test results being included in the PR so we can merge it faster and know what test to expect.

So I am thinking maybe you want to... revert the commit and redo it in a different way?

Many thanks. I do appreciate this PR.

soimort:
Non-video-service websites should be preferably handled by `embed.py`.

www.freebuf.com doesn't host anything interesting -- they're simply embedding videos from other sites. There are too many websites like this on the Internet, we can't maintain so many extractors for every of them -- and these could be fragile to changes.

Thanks, anyway.

soimort:
Plus, if a website seems to go down very easily, it's not worth it, IMHO.

screenshot from 2016-07-13 08-36-04

PR #1281, opened 2016-07-15
add qq and letv embed video support code

natugrey:
link`http://www.freebuf.com/video/108532.html`
return
```
Site:       QQ.com

  「安全大咖说」专访360核心安全负责人：郑文彬（MJ0011）| FB视频 - FreeBuf.COM | 关注黑客与极客
Size:       8.43 MiB (8840889 Bytes)
```
link`http://open.freebuf.com/inland/basic/520.html`
return
```
Site:       Le.com
Title:        什么是DDoS(分布式拒绝攻击)？|FreeBuf公开课 
Size:       26.98 MiB (28293854 Bytes)

```

### DIFF
--- a/src/you_get/extractors/embed.py
+++ b/src/you_get/extractors/embed.py
@@ -42,6 +42,14 @@ netease_embed_patterns = [ '(http://\w+\.163\.com/movie/[^\'"]+)' ]
 
 vimeo_embed_patters = [ 'player\.vimeo\.com/video/(\d+)' ]
 
+qq_embed_patterns = [ 'v\.qq\.com/iframe/player.html\?vid=([0-9a-zA-Z]+)' ]
+
+"""
+refer to http://help.lecloud.com/Wiki.jsp?page=PC4.0
+"""
+letv_embed_patterns =  [ 'http://yuntv.letv.com/player/vod/bcloud.js',
+                         'http://yuntv.letv.com/bcloud.js' ]
+
 
 def embed_download(url, output_dir = '.', merge = True, info_only = False ,**kwargs):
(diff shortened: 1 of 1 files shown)


src/you_get/extractors/embed.py
```diff
@@ -68,6 +76,17 @@ def embed_download(url, output_dir = '.', merge = True, info_only = False ,**kwa
         found = True
         iqiyi_download_by_vid((vid[1], vid[0]), title=title, output_dir=output_dir, merge=merge, info_only=info_only)
 
+    vids = matchall(content, qq_embed_patterns)
+    for vid in vids:
+        found = True
+        qq_download_by_vid(vid, title=title, output_dir=output_dir, merge=merge, info_only=info_only)
+
+    if len(matchall(content,letv_embed_patterns)):
```
soimort: What will happen if `matchall(content,letv_embed_patterns)` is `None`?

soimort: Although the return type of `matchall` can't be `None` (by definition), it would be better to make that bulletproof (just like all other matching patterns here).

zhangn1985: return value for matchall can be `None`.
just use `for ... in...` and it's fine.
don't use  `len` on `None`.

zhangn1985: and one more thing, `matchall` can match multiple match group.

eg: 
```
lecloud_embed_patterns = [ '{"uu":"([^\"]+)","vu":"([^\"]+)"',
                     ]
```
when match, `matchall` will return a list of `tuple`

PR #1282, opened 2016-07-16
bug fix for iqiyi.py not import json_output module

wwqgtxx:
bug fix for iqiyi.py not import json_output module

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1282)


src/you_get/extractors/mgtv.py
```diff
@@ -75,8 +75,9 @@ def prepare(self, **kwargs):
                 size_this_stream = 0
                 stream_fileid_list = []
                 for i in segment_list_this:
-                    _, container_this_stream, size_this_seg = url_info(i)
-                    size_this_stream += size_this_seg
+                    if 'json_output' not in kwargs and not kwargs['json_output']:
```
soimort: What does this do?

cnbeining: Why should VideoExtractor import this? Didn't the father VideoExtractor object fix the problem?

soimort:
@cnbeining  Please review this.

@wwqgtxx This PR looks fine, but the commits need to be squashed.

src/you_get/extractors/mgtv.py
soimort: `VideoExtractor` already takes care of `json_output`, it seems. So I need more information on this change.

soimort:
The `iqiyi.py` part looks good to me. So please split unrelated changes into different PRs (in the future), so that we can review and merge on an individual basis.
.

src/you_get/extractors/mgtv.py
cnbeining: @wwqgtxx We are not complaining about this PR: we are talking about a potential bug deeply hiding within EVERY ```VideoExtractor``` . If that is the case, we'd better fix it for it could be critical. Otherwise, can you kindly explain why you are trying to do this hacking? What unexpected behaviour are you having? 

Many thanks.

soimort: If `'json_output' not in kwargs`, trying to evaluate `not kwargs['json_output']` will raise an exception.
So this bool expression doesn't literally make sense.

@cnbeining Is it possible that `url_info` could take a long time so this PR is trying to get rid of that when retrieving the `--json` output?

cnbeining: I don't think so.

As I can remember MGTV's video URL is only **one** for each resolution(total resolution is 3 for most of the cases), backed by ChinaNet(WS): If the ```url_info``` fails, I can hardly imagine that the video can be downloaded by any chance.

cnbeining: I talked to the author: He DOES NOT want to get the information of every segment to reduce time. This is not what WE want here...

cnbeining:
Close, as author requested. Replaced by #1285 .

PR #1289, opened 2016-07-20
Fix Youku playlist download problem

hjk41:
Currently Youku playlist download is buggy, it downloads only the first page. Issue #1033 is opened for this problem.

The cause is that Youku's URL of the playlist page does not match what the script expects, so the script degrades to download only the first page.
This PR will fix the issue by modifying the URL and try to traverse all the pages.

You can try it with:
`you-get http://list.youku.com/albumlist/show?id=22238259&ascending=1&page=1`
and it should download all of the 50 videos.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1289)

### DIFF
--- a/src/you_get/extractors/youku.py
+++ b/src/you_get/extractors/youku.py
@@ -76,7 +76,7 @@ class Youku(VideoExtractor):
         for x in xs:
             if x not in mem:
                 mem.add(x)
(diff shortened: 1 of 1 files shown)


soimort:
Thanks again!

PR #1290, opened 2016-07-20
fix DNS resolve issue when using socks proxy

hjk41:
Some ISPs in mainland China now pollutes DNS for websites like tumblr and youtube. When using SOCKS proxy, we need to resolve the DNS through the SOCKS proxy, too.

This patch will enable resolving DNS through the specified SOCKS proxy, so it would work correctly.

To validate this patch, you can use a SOCKS5 proxy (127.0.0.1:1080 in this case) and then issue the following command:
`you-get -s 127.0.0.1:1080 http://muslimkings.tumblr.com/post/146985235471/me-when-a-straight-white-boy-tries-talking-2-me`

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1290)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -1214,15 +1214,18 @@ def script_main(script_name, download, download_playlist, **kwargs):
 
     if (socks_proxy):
         try:
-          import socket
-          import socks
-          socks_proxy_addrs = socks_proxy.split(':')
-          socks.set_default_proxy(socks.SOCKS5, 
-                                  socks_proxy_addrs[0], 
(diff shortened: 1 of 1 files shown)


cnbeining:
This makes sense.

I am just thinking whether we should make an argument for this feature, or we just override the DNS anyways.

Anyway, I cannot really test this PR, so I would leave it to someone in the Mainland. 

To anyone who is reading: Please report. Thanks.

我在墙外，测不出问题。

如果您在墙内，无论是否可用，请回报。感谢。

soimort:
Looks good. Thanks :+1:

PR #1309, opened 2016-07-27
[bilibili] video title may have space around

chideat:
There may be space around the finally file name. 
It's necessary for these who embed `you-get` in their program and get the title from stdout.

Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1309)


cnbeining:
GJ.

PR #1333, opened 2016-08-06
[Douyutv]Make correct did by uuid instead of random string

steven7851:
I'm not sure that lapi's 403 error is caused by random string. Just try to make correct did by uuid module.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1333)


soimort:
Any sample URL that leads to a 403 error?

steven7851:
@soimort It occur randomly, not specific channel.

PR #1335, opened 2016-08-08
Fix builtins.KeyError.

zoumi:
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1335)


soimort:
May I ask what problem does it solve?

zoumi:
I called the function: any_download() directly,so this may happen.(I didn't know there was a main_download() then.)

PR #1338, opened 2016-08-08
[Douyutv]fix channel on event

steven7851:
some channel on event will redirect to event page
exp. http://www.douyu.com/65251 redirect to http://www.douyu.com/cms/zt/RioOlympics.html#a
if channel is on event, use entering url to get room_id, and non-digital room_id will not work

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1338)


soimort:
👍

PR #1390, opened 2016-09-04
[universal] fix issue: "flush" is not available in Python under 3.3

Cappuccin0:
[universal] fix issue: "flush" is not available in Python under 3.3

Changes:
1. added a new function in common module named "print_more_compatible", which supports 'flush' keyword in pyver < 3.3 (e.g. 3.2.x), so the program won't be stopped by the unexpected exception. And in pyver >=3.3, it is identical as the built-in print function.
2. added "import" command in related files to overload the print function.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1390)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -1046,6 +1046,22 @@ def set_http_proxy(proxy):
     opener = request.build_opener(proxy_support)
     request.install_opener(opener)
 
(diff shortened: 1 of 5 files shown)


soimort:
Looks good, thanks 👍

PR #1393, opened 2016-09-07
Added support for panda.tv streams that do not use the standard plflag

Erk-:
Added support for panda.tv streams that do not use the standard plflag, it is mostly mobile streams that use the nonstandard plflags.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1393)

### DIFF
--- a/src/you_get/extractors/panda.py
+++ b/src/you_get/extractors/panda.py
@@ -18,10 +18,11 @@ def panda_download(url, output_dir = '.', merge = True, info_only = False, **kwa
     data = json.loads(content)['data']
     title = data.get('roominfo')['name']
     room_key = data.get('videoinfo')['room_key']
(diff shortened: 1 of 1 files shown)


soimort:
Any links for testing? Thanks.

Erk-:
It is changeing a bit what plflag a stream is using I think it has something to do with what it is streamed from. An example you can try now is "www.panda.tv/10009" it tries to play it with the url: "http://pl3.live.panda.tv/live_panda/057595ac05dfa5a2da4e849c3856f7cf.flv" when it should use the url: "http://pl8.live.panda.tv/live_panda/057595ac05dfa5a2da4e849c3856f7cf.flv"

soimort:
:+1:

Issue #1405, opened 2016-09-18
[Suggestion] Discard the the last 1.9MB video fragment from Youku


woodword:
Recently Youku automatically adds a 1.9 MB (1,883,240 bytes)  / 6 seconds video fragment behind the every video (at least for the videos I subscribed) uploaded by users. If you use "-n" / "--no-merge" switch you would find it's a commercial and quite annoying. 

What is worse is this fragment seems always be in 1920 x 1080 resolution, so if the video uploaded by the user is not in the same resolution it would lead to a merging failure. 

Please add some codes so that you-get could detect this fragment (md5 hash ?) and not download or try to merge it.

Some examples:
http://v.youku.com/v_show/id_XMTYyNjg5MTQwOA==.html
http://v.youku.com/v_show/id_XMTcyODYzMDQwOA==.html
http://v.youku.com/v_show/id_XMTYzNzg3NjkyNA==.html

---------------------------------------------------------------------------------------------------------------
UPDATE: After further test, it seems that the "fail to merge" error only happens to some old videos. For the newly uploaded ones Youku can choose the correct resolution for this 6-second commercial.

---------------------------------------------------------------------------------------------------------------
UPDATE: After I reinstalled the FFmpeg I can merge the 1920 x 1080 commercial, so it is only a suggestion for you-get now to discard the last 1.9MB video fragment. 
    This could be implemented by probing the duration of all downloaded fragments (ffprobe) and discarding those shorter than a few seconds, like 10 seconds, or, as @Erk- suggested,  ignoring all the videos fragments from certain addresses such as vali.cp31.ott.cibntv.net.

lamlam123tld3:
i dont get this problem er

woodword:
You mean you didn't get the last 6 seconds commercial , or you got it but you can merge it ?

Erk-:
The commercial comes from the url http://vali.cp31.ott.cibntv.net for me in both of your examples. I can merge it. Can you post your output of the "you-get --json http://v.youku.com/v_show/id_XMTcyODYzMDQwOA==.html".

Erk-:
I have created a fix but I am not sure it works on all urls #1407

lamlam123tld3:
for me,
i totally do not got this issues,
maybe becuz of my overseas IP?


a better method that may fix it
remove yxon  = 1 on line 307
to directly get the file link
instead of going through youku server



`q = parse.urlencode(dict(
                            ctype = self.ctype,
                            ev    = 1,
[...]

PR #1407, opened 2016-09-19
Fix for #1405

Erk-:
This is a fix for the short commercial in the end of the videos from youku it is not a pretty ix but fixed it in the examples give.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1407)


soimort:
Looks good, thanks 👍

PR #1424, opened 2016-09-27
fix bug about renaming videos when downloading with ffmpeg

Yohohaha:
when i was downloading videos at iqiyi, i found the video name can't be renamed, the problem is supposed to be when downloading with ffmpeg, it hasn't dealed with the output_filename variable, i have fixed it and now it works well.

---

This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1424)


Erk-:
Does this also fix #1409?

soimort:
Please fix the syntax error and make sure this works on your local machine.

Yohohaha:
It can work well on my pc, and it seems can fix #1409.

Erk-:
@Yohohaha  you need a ":" on the end of line 972 in common.py it shoud be
`if(output_filename):` instead of `if(output_filename)`

Yohohaha:
oh, thanks,

PR #1447, opened 2016-10-09
Fix http/https mixed content on index page

petrkle:
you-get.org is server over https so images should be secured to.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1447)


soimort:
Nice catch :+1:

PR #1448, opened 2016-10-10
translate char | to - in filename for ffmpeg concat

chideat:
translate char | to - in filename at all platform, for ffmpeg use char | to concat video slices.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1448)


soimort:
Looks good to me, thanks :+1:

PR #1472, opened 2016-10-20
icourse: add supprt

liushuyu:
Add support for iCourses

Should close #1451

**Known issues**
- Playlist download **ONLY** can originate from `http://www.icourses.cn/coursestatic/course_*.html` like sites
- Sometimes, the server will refuse the request due to high traffic (rush hour?)
- Sometimes, the request will fail due to poor network condition (but will retry)

---

This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1472)

PR #1478, opened 2016-10-25
common: add dynamic url support for `url_save_chunked`

liushuyu:
Add support for downloading the dynamic video URLs from some annoying site.

---

This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1478)


liushuyu:
This should be a pre-requirement for #1472

PR #1492, opened 2016-11-03
[embed] correct tudou pattern

zmwangx:
Hyphen-minus (`-`) is a valid character in Tudou's video ID. It's even present in the second pattern of `tudou_embed_patterns`, just not the first.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1492)


soimort:
Thanks!

PR #1511, opened 2016-11-13
[MGTV] Some fixes. Because it's not working

moolex:
1. Fix video url matcher
2. Change container from FLV to TS
3. Use m3u8 ext-info for fast size calculate
4. Use m3u8 url for video playing

Some test urls:

http://www.mgtv.com/b/298050/3690375.html
http://www.mgtv.com/b/294045/3348671.html

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1511)


soimort:
Thanks!

PR #1514, opened 2016-11-15
Support new mgtv url pattern: '/b/\d+/${vid}.html'.

StephenPCG:
Recently I find mgtv has a new url pattern, which ``you-get`` failed to recognize. Here is an example video page: http://www.mgtv.com/b/298442/3689404.html .

This PR adds support to the new url pattern.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1514)


soimort:
Already fixed by #1511 -- thanks, anyway.

PR #1540, opened 2016-12-03
Add proxy authorization support, fix #1539

### LINKED ISSUE #1539: any possibility of supporting private http proxy?
no where to input credentials atm

soimort:
Because why not. People trust this script. And we (probably) don't steal passwords.

Usage:
```
$ you-get -x username:password@host:port [url]
```

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1540)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -1080,11 +1080,25 @@ def parse_host(host):
     return (hostname, port)
 
 def set_proxy(proxy):
-    proxy_handler = request.ProxyHandler({
-        'http': '%s:%s' % proxy,
-        'https': '%s:%s' % proxy,
-    })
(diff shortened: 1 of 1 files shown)


soimort:
Still `http_error_407`. Give up.

soimort:
Drat, no wonder I  can't get `ProxyDigestAuthHandler` to work with squid. It's a bug of Python: (And it was there for 4 yrs)
https://bugs.python.org/issue16095

soimort:
After some packet capturing, I found that basic auth is implied even without a `ProxyBasicAuthHandler` (although not well-documented anywhere in Python's docs). If your proxy supports *basic authentication*, then `-x username:password@host:port` should just work.

So this PR is perfectly useless💩.

PR #1550, opened 2016-12-10
Yizhibo support

Erk-:
Support for yizhibo.com

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1550)

### DIFF
--- a/src/you_get/extractors/yizhibo.py
+++ b/src/you_get/extractors/yizhibo.py
@@ -9,7 +9,7 @@ import time
 def yizhibo_download(url, output_dir = '.', merge = True, info_only = False, **kwargs):
     video_id = url[url.rfind('/')+1:].split(".")[0]
     json_request_url = 'http://www.yizhibo.com/live/h5api/get_basic_live_info?scid={}'.format(video_id)
-    content = get_html(json_request_url)
+    content = get_content(json_request_url)
     error = json.loads(content)['result']
     if (error != 1):
         raise ValueError("Error : {}".format(error))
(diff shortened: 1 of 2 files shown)


SeaHOH:
http://www.mixcloud.com/DJVadim/north-america-are-you-ready/
The url in test must be insteaded or removed, it often returns `403 Forbidden`.
I am poor with English. How told the managers? It is always bot bot bo...

Erk-:
There is some problem with ffmpeg_download_stream not downloading video I will reopen it when I have fixed it.

Erk-:
I have fixed the problem by useing the download_urls instead of ffmpeg.

src/you_get/extractors/yizhibo.py
```diff
@@ -0,0 +1,37 @@
...
+    error = json.loads(content)['result']
+    if (error != 1):
+        raise ValueError("Error : {}".format(error))
+
+    data = json.loads(content)
+    title = data.get('data')['live_title']
+    if (title == ''):
+        title = data.get('data')['nickname']
+    m3u8_url = data.get('data')['play_url']
+    m3u8 = get_content(m3u8_url)
```
zhangn1985: are you sure to do m3u8 extract? for M3u8 live streaming, m3u8 link content only has current slides of the stream, maybe 1 or 2 mins or less.
could you show the result of real_url? and the video length?

SeaHOH:
No, I was said the url in unittest, not your code.  It makes checks failed.

src/you_get/extractors/yizhibo.py
Erk-: I could get the video length from the metadata in the m3u8 file but I do not belive there is an easy way to get the total size.
I tested the program on a 2 hour 30 minutes stream and it gave about 1000 .ts files. So there may be a better way.

I have not tested it proberly for livestreaming I will look at that today, it may only work for vods at the moment.

Erk-:
@SeaHOH I know but for some reason ffmpeg was not outputting the video stream.

soimort:
Thanks for the contribution:+1:

PR #1558, opened 2016-12-16
[ffmpeg] fix concat list when output dir is not pwd

zmwangx:
Relative paths in the concat list are considered relative to the parent directory of the script, not the calling directory. This isn't entirely obvious from the documentation, but it is easy to infer from the concat demuxer's concept of "safety", and easy to test (confirmed on FFmpeg 3.2.2). See https://ffmpeg.org/ffmpeg-all.html#concat-1 for details.

This commit fixes the wrong relative paths when `--output-dir` is specified and not pwd.

This commit also
- Factors out common concat list writer code;
- Slightly simplifies the code to collect FFmpeg params (on Py35+ we can further simplify by unpacking `LOGLEVEL` with the star operator right in the list literal).

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1558)

### DIFF
--- a/src/you_get/processor/ffmpeg.py
+++ b/src/you_get/processor/ffmpeg.py
@@ -26,6 +26,18 @@ LOGLEVEL = ['-loglevel', 'quiet']
 def has_ffmpeg_installed():
     return FFMPEG is not None
 
+# Given a list of segments and the output path, generates the concat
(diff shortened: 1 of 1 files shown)


soimort:
LGTM. Thanks!:+1:

PR #1559, opened 2016-12-16
Fix for magisto

Erk-:
Use the api instead of finding the source on the page.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1559)


soimort:
Cool. Thanks👍

PR #1577, opened 2016-12-26
Update pptv.py

redstoneleo:
fixed pptv video url match

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1577)

### DIFF
--- a/src/you_get/extractors/tudou.py
+++ b/src/you_get/extractors/tudou.py
@@ -13,8 +13,9 @@ def tudou_download_by_iid(iid, title, output_dir = '.', merge = True, info_only 
 
     urls = []
     for vid in vids:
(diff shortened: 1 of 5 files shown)


src/you_get/extractors/tudou.py
```diff
@@ -13,8 +13,9 @@ def tudou_download_by_iid(iid, title, output_dir = '.', merge = True, info_only
 
     urls = []
     for vid in vids:
-        for i in parseString(get_html('http://ct.v2.tudou.com/f?id=%s' % vid)).getElementsByTagName('f'):
-            urls.append(i.firstChild.nodeValue.strip())
+        urls.append('http://vr.tudou.com/v2proxy/v?sid=95000&id={}&st=99'.format(vid))
```
soimort: What does this commit do?

redstoneleo: do the same thing as the above two lines do

PR #1581, opened 2016-12-29
Fix soku.com vid download error problem

liujianshan:
修复youku用vid下载报错导致不能下载列表所有视频的问题

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1581)


soimort:
Thanks for the fix! 👍

Issue #1569, opened 2016-12-21
bilibili 番剧解析问题


mengmo:
```
you-get -di http://bangumi.bilibili.com/anime/427/play#71958
[DEBUG] get_content: http://bangumi.bilibili.com/anime/427/play#71958
[DEBUG] post_content: http://bangumi.bilibili.com/web_api/get_source
 post_data: {'episode_id': None}
you-get: version 0.4.595, a tiny downloader that scrapes the web.
you-get: ['http://bangumi.bilibili.com/anime/427/play#71958']
Traceback (most recent call last):
  File "<string>", line 11, in <module>
  File "src\you_get\__main__.py", line 92, in main
  File "src\you_get\common.py", line 1400, in main
  File "src\you_get\common.py", line 1313, in script_main
  File "src\you_get\common.py", line 1129, in download_main
  File "src\you_get\common.py", line 1393, in any_download
  File "src\you_get\extractors\bilibili.py", line 133, in bilibili_download
KeyError: 'result'
you-get returned -1
```
B站`bangumi`的`URL格式`改了，原来是这样的
```
you-get -di http://bangumi.bilibili.com/anime/v/71958
[DEBUG] get_content: http://bangumi.bilibili.com/anime/v/71958
[DEBUG] post_content: http://bangumi.bilibili.com/web_api/get_source
 post_data: {'episode_id': None}
you-get: version 0.4.595, a tiny downloader that scrapes the web.
you-get: ['http://bangumi.bilibili.com/anime/v/71958']
Traceback (most recent call last):
  File "<string>", line 11, in <module>
  File "src\you_get\__main__.py", line 92, in main
  File "src\you_get\common.py", line 1400, in main
  File "src\you_get\common.py", line 1313, in script_main
  File "src\you_get\common.py", line 1129, in download_main
  File "src\you_get\common.py", line 1393, in any_download
  File "src\you_get\extractors\bilibili.py", line 133, in bilibili_download
KeyError: 'result'
you-get returned -1
```
对应av号的`URL格式`也不行
```
you-get -di https://www.bilibili.com/video/av2983852/
[DEBUG] get_content: http://www.bilibili.com/video/av2983852/
you-get: version 0.4.595, a tiny downloader that scrapes the web.
you-get: ['https://www.bilibili.com/video/av2983852/']
Traceback (most recent call last):
  File "<string>", line 11, in <module>
  File "src\you_get\__main__.py", line 92, in main
[...]

PR #1586, opened 2016-12-31
fix bilibili bangumi support

zjuchenyuan:
modify the regex to get eposide id

bilibili has changed their html source code, change the regex would help fix issue #1569

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1586)


soimort:
Looks good, thanks!

PR #1665, opened 2017-02-02
[ffmpeg] call ffmpeg with stdin redirected to the null device

zmwangx:
Prevent FFmpeg from consuming stdin and interpreting the character stream as a stream of interactive commands, specifically:

    ?      show this help
    +      increase verbosity
    -      decrease verbosity
    c      Send command to first matching filter supporting it
    C      Send/Queue command to all matching filters
    D      cycle through available debug modes
    h      dump packets/hex press to cycle through the 3 states
    q      quit
    s      Show QP histogram

This is mostly useful in a batch environment where an unsuspecting user
may not realize you-get could consume stdin through FFmpeg, e.g.

```sh
while read url; do you-get "$url"; done <urllist
```

(Related: BashFAQ/089.)

Yeah, apparently the above could easily be replaced with `xargs you-get <urllist` which is more elegant and doesn't suffer from the stdin consumption problem, but the above is a pretty honest and straightforward command line which could been seen in the real world.

In addition, even if we ignore the contrived command line, a user probably won't be thrilled if he accidentally smashes the "h" key and all of a sudden megabytes worth of dumps pour in on stderr.

---

An alternative would be FFmpeg's `-nostdin` option, which explicitly disables interaction, but adding such an option to existing code would be even more destructive than the `stdin=DEVNULL` approach taken here.

---

This commit also happens to fix a dual-call bug in `ffmpeg_concat_mp4_to_mpg`.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1665)

### DIFF
--- a/src/you_get/processor/ffmpeg.py
+++ b/src/you_get/processor/ffmpeg.py
@@ -6,9 +6,18 @@ import subprocess
 from ..util.strings import parameterize
 from ..common import print_more_compatible as print
 
+try:
+    from subprocess import DEVNULL
+except ImportError:
+    # Python 3.2 or below
+    import os
+    import atexit
+    DEVNULL = os.open(os.devnull, os.O_RDWR)
+    atexit.register(lambda fd: os.close(fd), DEVNULL)
+
 def get_usable_ffmpeg(cmd):
     try:
-        p = subprocess.Popen([cmd, '-version'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        p = subprocess.Popen([cmd, '-version'], stdin=DEVNULL, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         out, err = p.communicate()
         vers = str(out, 'utf-8').split('\n')[0].split()
         assert (vers[0] == 'ffmpeg' and vers[2][0] > '0') or (vers[0] == 'avconv')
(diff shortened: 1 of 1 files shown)


src/you_get/processor/ffmpeg.py
```diff
@@ -6,9 +6,18 @@
 from ..util.strings import parameterize
 from ..common import print_more_compatible as print
 
+try:
+    from subprocess import DEVNULL
+except ImportError:
+    # Python 3.2 or below
```
zmwangx: Not sure if we need to maintain Python 3.2 compatibility. If not, this compat block can be removed.

src/you_get/processor/ffmpeg.py
```diff
@@ -92,9 +101,8 @@ def ffmpeg_concat_mp4_to_mpg(files, output='output.mpg'):
     params.append(output + '.mpg')
     params += ['-vcodec', 'copy', '-acodec', 'copy']
     params.append(output)
-    subprocess.call(params)
 
-    if subprocess.call(params) == 0:
+    if subprocess.call(params, stdin=DEVNULL) == 0:
```
zmwangx: Here is the dual-call bug I was talking about. I didn't read surrounding code but it's safe to assume that `subprocess.call(params)` shouldn't be called twice in immediate succession.

soimort:
Actually it would be desirable if users *can* use these interactive  keys as an expected behavior, since they're standard to ffmpeg.

Perhaps allow `stdin` if `you-get --debug` is specified?

zmwangx:
> Actually it would be desirable if users can use these interactive keys as an expected behavior, since they're standard to ffmpeg.

One concern would be lack of consistency, since `ffmpeg_download_stream`, for instance, connects FFmpeg's stdin to a pipe (so that the program could use the `q` command to stop recording). But allowing interactive commands in debug mode seems like a reasonable middle ground. I pushed an update.

soimort:
LGTM 👍

PR #1676, opened 2017-02-09
[nicovideo] fix extractor

zmwangx:
- Rework login. The existing login system appears to be dysfunctional; the cookiejar after a login attempt with the existing `nicovideo_login` function looks like (cookie values have been scrubbed and replaced with xxx):

      <CookieJar[<Cookie mail_for_retry=xxx for .nicovideo.jp/>, <Cookie nicosid=xxx for .nicovideo.jp/>]>

  and with these cookies, the `getflv` API simply returns

      closed=1&done=true

  Using the refreshed `nicovideo_login` function, the cookiejar after a login attempt looks like:

      <CookieJar[<Cookie nicosid=xxx for .nicovideo.jp/>, <Cookie user_session=user_session_xxx for .nicovideo.jp/>, <Cookie user_session_secure=xxx for .nicovideo.jp/>]>

  and the `getflv` API functions correctly with these cookies.

- Make title extraction more robust. Nicovideo.jp seems to serve up two different layouts at the moment, one of them being a beta version. See https://gist.github.com/ed95394afc6eff8a781395ac5afcbe48 for a sample page http://www.nicovideo.jp/watch/sm22221659 in both layouts. (Disclosure: I'm located in U.S., where I'm being served one of the two layouts at random across consecutive runs; YMMV elsewhere.) The title is embedded differently in different layouts.

  This commit makes sure both layouts are accounted for.

- http://www.nicovideo.jp/api/getflv is now 301 redirected to http://flapi.nicovideo.jp/api/getflv, so we switch to the new location.

- Switch from deprecated `get_html` to `get_content`.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1676)

### DIFF
--- a/src/you_get/extractors/nicovideo.py
+++ b/src/you_get/extractors/nicovideo.py
@@ -5,8 +5,15 @@ __all__ = ['nicovideo_download']
 from ..common import *
 
 def nicovideo_login(user, password):
-    data = "current_form=login&mail=" + user +"&password=" + password + "&login_submit=Log+In"
-    response = request.urlopen(request.Request("https://secure.nicovideo.jp/secure/login?site=niconico", headers=fake_headers, data=data.encode('utf-8')))
+    post_data = {
+        'mail_tel': user,
+        'password': password,
+    }
+    response = request.urlopen(request.Request(
+        'https://account.nicovideo.jp/api/v1/login?site=niconico',
+        headers=fake_headers,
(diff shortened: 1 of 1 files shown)


zmwangx:
Actually, the existing login API should still work; I just realized the problem was that it didn't percent encode my password.

Anyway, it doesn't really hurt to upgrade to the current login API used by https://account.nicovideo.jp/login.

PR #1677, opened 2017-02-09
[common] log URLs in more functions with network requests

zmwangx:
This is a follow-up to #999.

This commit adds the

    <function_name>: <url>

debug message, which was previously only emitted by `get_content` and `post_content`, to all high level utility functions with network requests except `url_size`, `url_save` and `url_save_chunked` (in order not to ruin progress bars).

---

Rationale: I've been very happy with the `get_content: <url>` messages since their introduction, so expansion to all utility functions seems a good idea to me.

To respond to the original review comment in #999:

> Honestly speaking, I've not needed debug messages like this; leaving tcpdump open and you'll get a bit of everything underhood (at least for HTTP-only sites like Youku and Tudou, which is commonly the case for most Chinese websites).
>
> However, it could be easier to tell in which URL you're stuck purely from the debug message of you-get itself. And it works for HTTPS links.

tcpdump works, but it could be very noisy when you're multitasking on one interface. Moreover, if you're running multiple you-get sessions at once, and maybe even have pages from the same website(s) open in the browser at the same time, it's hard to tell which request originated from which session.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1677)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -255,6 +255,8 @@ def undeflate(data):
 
 # DEPRECATED in favor of get_content()
 def get_response(url, faker = False):
+    logging.debug('get_response: %s' % url)
+
     # install cookies
     if cookies:
         opener = request.build_opener(request.HTTPCookieProcessor(cookies))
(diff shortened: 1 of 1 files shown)


soimort:
The more debug messages the better 👍

PR #1699, opened 2017-02-20
Print audiolang in json output

coslyk:
Print "audiolang" in json output if it is available, e.g. multi-languagge videos in youku.com


soimort:
👍

PR #1744, opened 2017-03-09
youku: warn about segments skipped due to paywall

zmwangx:
This is especially helpful in cases where the entire video is blocked by paywall, resulting in an unhelpful error message

```
you-get: [Failed] Cannot extract video source.
```

I spent quite a few minutes debugging a URL like that,^ which could have been saved by the message

```
you-get: Skipping 27 out of 27 segments due to paywall
```

added by this commit.

^The video is completely region-blocked where I am, so I didn't know it was behind a paywall before feeding it to you-get — with a Mainland-based extractor proxy of course. That should explain my initial confusion — it would have been obvious if I checked its availability in a Unblock Youku-enabled browser session first. Still, more self-contained information should make you-get better.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1744)

### DIFF
--- a/src/you_get/extractors/youku.py
+++ b/src/you_get/extractors/youku.py
@@ -295,9 +295,14 @@ class Youku(VideoExtractor):
                 for piece in pieces:
                     segs = piece['segs']
                     streamfileid = piece['fileid']
-                    for no in range(0, len(segs)):
+                    seg_count = len(segs)
+                    for no in range(0, seg_count):
                         k = segs[no]['key']
-                        if k == -1: break # we hit the paywall; stop here
+                        if k == -1:
+                            # we hit the paywall; stop here
+                            log.w('Skipping %d out of %d segments due to paywall' %
+                                  (seg_count - no, seg_count))
+                            break
                         fileid, ep = self.__class__.generate_ep(self, no, streamfileid,
                                                                 sid, token)
                         q = parse.urlencode(dict(


rosynirvana:
IMO the original url should be printed. Number of segments skipped does not help the user.

zmwangx:
Not sure what you mean. This is just a warning, essentially telling the user "look, there are this many segments we can't do anything about". All intermediate URLs are useless without keys.

rosynirvana:
Original urls are those urls from input of users.
Users should not be informed with segments.

zmwangx:
1. It's rather pointless to echo user input. Echoing back the URL is at best slightly more clear when users are downloading multiple videos, but even then due to the sequential nature of you-get, it's extremely easy to know exactly which URL the warning applies to.

    Note that currently error messages don't contain original URLs either. Not following conventions makes the (frankly already quite inconsistent, due to its nature) project inconsistent.

2. > Users should not be informed with segments.

    Users are already exposed to the total number of segments (that are not skipped) and the segment currently being downloaded in the progress bar. So this assertion has no ground.

    > Number of segments skipped does not help the user.

    As a user myself, it totally helps. When a video claimed to be 1GB (in fact, that number isn't available across the board) ended up with only 3 segments totally 100MB, knowing that 27 segments were skipped due to paywall instantly makes the situation clear, and helps estimate how much I'm missing out.

rosynirvana:
You mean the patch does not try to inform users which url has not been downloaded completely, but to tell users what percent of a single video file is downloaded?

zmwangx:
> You mean the patch does not try to inform users which url has not been downloaded completely

Again, this is obvious from context.

soimort:
Any sample URLs for testing?

zmwangx:
Sure, plenty of examples can be found on vip.youku.com. Just find a VIP-only episode of a TV show, e.g.,

```
$ you-get --debug https://v.youku.com/v_show/id_XMjY0Mzk0Mzk2MA==.html
[DEBUG] get_content: http://play.youku.com/play/get.json?vid=XMjY0Mzk0Mzk2MA==&ct=10
[DEBUG] get_content: http://play.youku.com/play/get.json?vid=XMjY0Mzk0Mzk2MA==&ct=12
you-get: Skipping 24 out of 24 segments due to paywall
site:                优酷 (Youku)
title:               热血长安 第一季 24 千年太岁
stream:
    - format:        hd3
      container:     flv
      video-profile: 1080P
      size:          1209.7 MiB (1268423229 bytes)
    # download-with: you-get --format=hd3 [URL]

audio-languages:
    - lang:          国语
      download-url:  http://v.youku.com/v_show/id_XMjY0Mzk0Mzk2MA==

you-get: [Failed] Cannot extract video source.
```

soimort:
👍

PR #1745, opened 2017-03-09
log: mark xterm* terminals as ANSI escape sequences-compatible

zmwangx:
`xterm-color`, `xterm-16color`, `xterm-88color` and `xterm-256color` are now covered.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1745)

### DIFF
--- a/src/you_get/util/log.py
+++ b/src/you_get/util/log.py
@@ -5,13 +5,13 @@ from ..version import script_name
 
 import os, sys
 
-IS_ANSI_TERMINAL = os.getenv('TERM') in (
+TERM = os.getenv('TERM', '')
+IS_ANSI_TERMINAL = TERM in (
     'eterm-color',
     'linux',
     'screen',
     'vt100',
-    'xterm',
-)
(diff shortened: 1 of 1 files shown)


rosynirvana:
```
IS_ANSI_TERMINAL = os.getenv('TERM') in (
    'eterm-color',
    'linux',
    'screen',
    'vt100',
) or os.getenv('TERM').startswith('xterm')
```
If there's not a TERM env the code will raise an AttributeError

```os.getenv('TERM')``` could be None. IMHO Checking membership of None is just absurd and should be avoided. An empty set is a subset of any set but python doesn't following this, which could lead to confusion.

zmwangx:
Thanks, fixed.

zmwangx:
By the way, checking membership of `None` is perfectly fine. (1) An empty sequence or set type is vastly different from `None`; (2) `in` is a membership test operator; it doesn't test subset/subsequence.

soimort:
Thanks!

PR #1776, opened 2017-03-18
Fix json output

coslyk:
Change "NameError" to "AttributeError" in json_output.py

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1776)


soimort:
Thanks again!

PR #1788, opened 2017-03-24
[Bilibili] Workaround for speed limiting issue (#1784)

soimort:
1. Added an additional `timeout` parameter in `url_save()`.
2. Once detected a `socket.timeout`, the Bilibili extractor should reinitialize the download.

---
This change is [Reviewable](https://reviewable.io/reviews/soimort/you-get/1788)


rosynirvana:
There is a chance that socket timeout doesn't work. 
Downloading n MiB -> reconnect -> Downloading n MiB could be better.

soimort:
We'll see if there's a better workaround _(:3」∠)_

PR #1852, opened 2017-04-15
fix bug  -- youku change the JSON metadata format

xuhaomin:
fix the bug , youku change the JSON metadate format from http://play.youku.com/play/get.json?
for example : http://play.youku.com/play/get.json?vid=XMjcwOTIwMzAwOA&ct=12  response json  use new fileid at [data][stream][segs][*][fileid] instead old streamfileid at  [data][stream][stream_fileid]


xuhaomin:
fix the bug , youku change the JSON metadate format

Chigogo:
Win exe用户该怎么用呀……

soimort:
Fixed by #1857. Thanks for the PR!

PR #1884, opened 2017-04-20
Tuning debugging message

rosynirvana:
1. print stack trace when ```UnicodeEncodeError``` caught
2. remove deugging message from ```get_html``` and ```get_decoded_html```, which call ```get_responce``` and print two message for one HTTP request.
3. print a hit when an ffmpeg nightly build, which uses concat protocol when merges video segs.


soimort:
Nice, thanks.

Issue #1908, opened 2017-04-25
List of QQ Videos Don't Download


dkl3:
Please make sure these boxes are checked before submitting your issue – thank you!

- [x] You can actually watch the video in your browser or mobile application, but not download them with `you-get`.
- [x] Your `you-get` is up-to-date.
- [x] I have read <https://github.com/soimort/you-get/wiki/FAQ> and tried to do so.
- [x] The issue is not yet reported on <https://github.com/soimort/you-get/issues> or  <https://github.com/soimort/you-get/wiki/Known-Bugs>. If so, please add your comments under the existing issue.
- [x] The issue (or question) is really about `you-get`, not about some other code or project.

Run the command with the `--debug` option, and paste the full output inside the fences:

```
$ you-get -I list --debug
[DEBUG] get_response: http://v.qq.com/page/a/0/g/a01714fzv0g.html
you-get: version 0.4.715, a tiny downloader that scrapes the web.
[...]

PR #1909, opened 2017-04-26
[qq]fix v.qq.com/page redirect

rosynirvana:
Now some ```v.qq.com/page``` redirected by HTTP 302: #1908

test case
```
./you-get -di 'http://v.qq.com/page/a/0/g/a01714fzv0g.html'
[DEBUG] url_locations: http://v.qq.com/page/a/0/g/a01714fzv0g.html
[DEBUG] get_response: https://v.qq.com/x/page/a01714fzv0g.html
[DEBUG] get_response: http://vv.video.qq.com/getinfo?otype=json&appver=3%2E2%2E19%2E333&platform=11&defnpayver=1&vid=a01714fzv0g
[DEBUG] get_response: http://vv.video.qq.com/getkey?otype=json&platform=11&format=10701&vid=a01714fzv0g&filename=a01714fzv0g.p701.1.mp4
[DEBUG] url_info: http://42.236.95.151/vhot2.qqvideo.tc.qq.com/a01714fzv0g.p701.1.mp4?vkey=51E23FF9507E1D542F1FDC3D41C1274C7AD8FA110C319E3B805B9977B58E3F498E2C8EB7D8D9AB27338FFE18C51DF81BA483EB18F146B8A35B9A269F183BE3E744224776151DFB328F9A76B0E925C3D286BB749406C5B96E83EC642850AE33B0
[DEBUG] get_response: http://vv.video.qq.com/getkey?otype=json&platform=11&format=10701&vid=a01714fzv0g&filename=a01714fzv0g.p701.2.mp4
Site:       QQ.com
Title:      中日艺术家同台演绎《舞蹈交响曲》 -CCTV大富
Size:       42.77 MiB (44852602 Bytes)
```

### DIFF
--- a/src/you_get/extractors/qq.py
+++ b/src/you_get/extractors/qq.py
@@ -134,8 +134,13 @@ def qq_download(url, output_dir='.', merge=True, info_only=False, **kwargs):
     if 'v.qq.com/page' in url:
         # for URLs like this:
         # http://v.qq.com/page/k/9/7/k0194pwgw97.html
-        content = get_html(url)
-        url = match1(content,r'window\.location\.href="(.*?)"')
+        new_url = url_locations([url])[0]
+        if url == new_url:
+            #redirect in js?
+            content = get_content(url)
+            url = match1(content,r'window\.location\.href="(.*?)"')
+        else:
+            url = new_url
 
     if 'kuaibao.qq.com' in url or re.match(r'http://daxue.qq.com/content/content/id/\d+', url):
         content = get_html(url)


soimort:
👍

dkl3:
ok i'm confused. i tried downloading the two videos again, and one still overwrites the other (e.g. FIRE EMBLEM Echoes 另一位英雄王)

rosynirvana:
This patch only fixes redirects.
There's little we can do about identical names with current code.

dkl3:
what about adding a message that says "filename already exists--replace or rename file?

rosynirvana:
A fix is possible, but could mess up too much.
The problem is download resumption heavily relies on identical names.
And by convention a commandline program silent overwrite the file with same name, unless the existed file is read-only.

dkl3:
Sorry for late reply, anyway

How about an option of adding the videos' identifier/token at the end of the filename like youtube-dl does?

rosynirvana:
I've no idea what youtube-dl does. FYI，an identifier(a hash?) is not always possible to get and I can't see how a token helps.

Handling identical filename is not trivial. IMO the correct way to impl it should check all the input first. But now ```you-get``` has at least 3 ways to handle multi-input and it's not quite possible to unify. In short, it's now quite posible with current code.

IMO you can try to fix it yourself because you know what you need and what you don't need. The fix wil be much easier.

Issue #1897, opened 2017-04-24
bilibili会下载两种清晰度的视频


7654:
Please make sure these boxes are checked before submitting your issue – thank you!

- [ ] You can actually watch the video in your browser or mobile application, but not download them with `you-get`.
- [x] Your `you-get` is up-to-date.
- [x] I have read <https://github.com/soimort/you-get/wiki/FAQ> and tried to do so.
- [x] The issue is not yet reported on <https://github.com/soimort/you-get/issues> or  <https://github.com/soimort/you-get/wiki/Known-Bugs>. If so, please add your comments under the existing issue.
- [x] The issue (or question) is really about `you-get`, not about some other code or project.

Run the command with the `--debug` option, and paste the full output inside the fences:

```
you-get http://www.bilibili.com/video/av9954010/
Site:       bilibili.com
Title:      【西瓜新】一个小小的八音盒【有一点花絮】
Type:       Flash video (video/x-flv)
Size:       41.05 MiB (43044992 Bytes)

Downloading 【西瓜新】一个小小的八音盒【有一点花絮】.flv ...
24.4% ( 10.0/ 41.1MB) ├██████████───────────────────────────────┤[1/1]    3 MB/s
Site:       bilibili.com
Title:      【西瓜新】一个小小的八音盒【有一点花絮】
Type:       Flash video (video/x-flv)
Size:       41.05 MiB (43044992 Bytes)

Downloading 【西瓜新】一个小小的八音盒【有一点花絮】.flv ...
48.7% ( 20.0/ 41.1MB) ├████████████████████─────────────────────┤[1/1]    4 MB/s
Site:       bilibili.com
Title:      【西瓜新】一个小小的八音盒【有一点花絮】
Type:       MPEG-4 video (video/mp4)
Size:       25.6 MiB (26847580 Bytes)

Downloading 【西瓜新】一个小小的八音盒【有一点花絮】.mp4 ...
39.1% ( 10.0/ 25.6MB) ├████████████████─────────────────────────┤[1/1]    5 MB/s
Site:       bilibili.com
Title:      【西瓜新】一个小小的八音盒【有一点花絮】
Type:       MPEG-4 video (video/mp4)
Size:       25.6 MiB (26847580 Bytes)

Downloading 【西瓜新】一个小小的八音盒【有一点花絮】.mp4 ...
78.1% ( 20.0/ 25.6MB) ├████████████████████████████████─────────┤[1/1]    6 MB/s
Site:       bilibili.com
[...]

PR #1922, opened 2017-05-02
[Bilibili] Add header when accessing video link

cnbeining:
Can someone in the Mainland test this for me? @rosynirvana your opinion would be welcomed.

If the speed limit IS due to the Referer, then we already support a quick bypass;

If the speed limit is due to the IP frequency, then there's little we can do from our side. 

For oversea users: Unblock-Youku and similar services are bounded by IP frequency regarding the speed limit. There's really little we can do from this project.

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -11,6 +11,16 @@ from .youku import youku_download_by_vid
 import hashlib
 import re
 
+FAKE_HEADER = {
+    'Accept-Encoding': 'gzip, deflate, br',
+    'Accept-Language': 'en-CA,en;q=0.8,en-US;q=0.6,zh-CN;q=0.4,zh;q=0.2',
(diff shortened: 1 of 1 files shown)


steven7851:
Maybe oversea users can use other API like this?
[bilibili_bangumi_area_limit_hack.user.js](https://github.com/ipcjs/bilibili-helper/blob/user.js/bilibili_bangumi_area_limit_hack.user.js)

cnbeining:
This project is designed to work without any 3rd party service.

One is free to make their own changes(and even raise PR), but the change shall not be merged.

rosynirvana:
Adding a referer works but IMO #1897 should be fixed in the next merge.

cnbeining:
@rosynirvana I personally would like to remove the auto-retry function within the code: too hacky IMHO and is adding unnecessary stress to Bilibili's CDN which is probably not what we want to do.

cnbeining:
@rosynirvana My mistake of not carefully reading through the whole issue history and your commit log. I will probably abandon this PR since this fix is within your PR. Let's move the discussion under your PR.

PR #1939, opened 2017-05-06
strip name of album fixed when download some album which name endwith…

bigfuji:
When download some album which name endwith spaces will rasie FileNotFoundError.
Because os.mkdir will automatically remove spaces in the end,so it will not find the correct path of tempfile.
Try this:http://music.163.com/#/album?id=34839428
当下载一些专辑名以空格结尾的专辑时，会引发FileNotFoundError。
因为os.mkdir会自动去除结尾的空格，所以程序找不到正确的临时文件的路径。
举个栗子：http://music.163.com/#/album?id=34839428


soimort:
Thanks:+1:

Issue #1942, opened 2017-05-06
bugs in common.py


rosynirvana:
```py
def url_save(url, filepath, bar, refer = None, is_part = False, faker = False, headers = {}, timeout = None, **kwargs):
    file_size = url_size(url, faker = faker, headers = headers)
```

If refer is set, it will be ignored in calling url_size.

soimort:
`url_size()` doesn't take `refer` as a parameter. I guess 683a475904e373db9b04d2a06d595a1818d5bbb0 is responsible for this...

PR #1945, opened 2017-05-06
[common] url_info should not ignore refer when called in url_save

rosynirvana:
#1942

details are in the following discussion.

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -479,6 +479,9 @@ def url_locations(urls, faker = False, headers = {}):
     return locations
 
 def url_save(url, filepath, bar, refer = None, is_part = False, faker = False, headers = {}, timeout = None, **kwargs):
+#When a referer specified with param refer, the key must be 'Referer' for the hack here
+    if refer is not None:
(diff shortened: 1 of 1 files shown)


soimort:
Why not:

```py
if refer:
    headers['referer'] = refer
file_size = url_size(url, faker = faker, headers = headers)
```

If `refer` is specified, it will be preferred rather than `headers['referer']` anyway. There's no need to keep the old value.

rosynirvana:
keys in headers are case insenstive by RFC so it can be Referer, REFERER.

soimort:
`headers` are provided by our side. Sticking with one convention (say, `headers['Referer']`) is fair enough. Moreover, `headers` will be used over and over again ([common.py#L529-534](https://github.com/soimort/you-get/blob/058d2797c5a8e1221cbe30e2a7240918b5b01c38/src/you_get/common.py#L529_L534)), while `new_headers` is just a one-time hack (simply works for `url_size()`).

rosynirvana:
I see.

soimort:
👌

PR #1946, opened 2017-05-06
Ku6/fix extension error

XiaochenCui:
Fix error when parse url "http://v.ku6.com/show/dBKZbnO51PWi7N2xRa3Bmw...html".
Case the sub-url end with mp4?AccessKeyId=...&Expires=...&Signature=...
So the regex in ku6 should be improved to fix the error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,8 +59,6 @@ docs/_build/
 target/
 
 # Misc
-_*
-*_
 *.3gp
 *.asf
 *.download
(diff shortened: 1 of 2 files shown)


XiaochenCui:
And I don't know if the changes in gitignore is necessary, but the rule _* makes __init__.py and __main__.py been ignored by git, which shouldn't been ignored obviously.

PR #1947, opened 2017-05-06
[ku6] Fix error when parse url's extension

XiaochenCui:
Fix error when parse url http://v.ku6.com/show/dBKZbnO51PWi7N2xRa3Bmw...html
Case the sub-url end with mp4?AccessKeyId=...&Expires=...&Signature=...
So the regex in ku6 should be improved to fix the error.


XiaochenCui:
With the original code, ext will be `mp4?AccessKeyId...` and download will abort.

soimort:
Nice .Thanks!

Issue #500, opened 2015-03-05
新浪视频UnicodeDecodeError


Geew:
URL: http://video.sina.com.cn/news/spj/topvideoes20150304/?opsubject_id=top1#137380095  

File "/usr/local/lib/python3.2/dist-packages/you_get-0.3.32-py3.2.egg/you_get/**main**.py", line 91, in main
    main()
  File "/usr/local/lib/python3.2/dist-packages/you_get-0.3.32-py3.2.egg/you_get/common.py", line 999, in main
    script_main('you-get', any_download, any_download_playlist)
  File "/usr/local/lib/python3.2/dist-packages/you_get-0.3.32-py3.2.egg/you_get/common.py", line 893, in script_main
    download_main(download, download_playlist, args, playlist, output_dir=output_dir, merge=merge, info_only=info_only)
  File "/usr/local/lib/python3.2/dist-packages/you_get-0.3.32-py3.2.egg/you_get/common.py", line 785, in download_main
    download(url, *_kwargs)
  File "/usr/local/lib/python3.2/dist-packages/you_get-0.3.32-py3.2.egg/you_get/common.py", line 992, in any_download
    m.download(url, *_kwargs)
  File "/usr/local/lib/python3.2/dist-packages/you_get-0.3.32-py3.2.egg/you_get/extractors/sina.py", line 67, in sina_download
    video_page = get_content(url)
  File "/usr/local/lib/python3.2/dist-packages/you_get-0.3.32-py3.2.egg/you_get/common.py", line 191, in get_content
    data = data.decode(charset)
UnicodeDecodeError: 'gb2312' codec can't decode bytes in position 54627-54628: illegal multibyte sequence

Geew:
get_content 函数 decode使用: data = data.decode('GB18030')就行了, 但是下面接下来的请求会报超时错误:


Traceback (most recent call last):
  File "you-get", line 19, in <module>
    you_get.main(repo_path=_filepath)
  File "/home/gowin/project/videoGet/libGit/you-get/src/you_get/__main__.py", line 91, in main
    main()
  File "/home/gowin/project/videoGet/libGit/you-get/src/you_get/common.py", line 1005, in main
    script_main('you-get', any_download, any_download_playlist)
  File "/home/gowin/project/videoGet/libGit/you-get/src/you_get/common.py", line 899, in script_main
    download_main(download, download_playlist, args, playlist, output_dir=output_dir, merge=merge, info_only=info_only)
[...]

PR #1982, opened 2017-05-17
[sina]try to find vid from url

rosynirvana:
#500

```
./you-get -i 'http://video.sina.com.cn/news/spj/topvideoes20150304/?opsubject_id=top1#137380095'
Site:       Sina.com
Title:      惠州博罗：红火蚁围村 村民谈蚁色变
Type:       Flash video (video/x-flv)
Size:       24.66 MiB (25858597 Bytes)
```

better than nothing?


soimort:
👍

PR #1698, opened 2017-02-20
bilibili multi fixes

rosynirvana:
~~在common.py里加了get_content_and_redirected这样一个函数，以tuple返回content和redirected_url~~

1. 用url_locations处理redirects，修正了#1693
test case
```
./you-get -i 'http://www.bilibili.com/video/av7873772/'
Site:       bilibili.com
Title:      URARA迷路帖 [1 少女与占卜、偶尔的露肚皮]
Type:       Flash video (video/x-flv)
Size:       345.69 MiB (362486000 Bytes)
```

2. 正确处理bangumi的episode titles
test case
```
./you-get -i 'http://bangumi.bilibili.com/anime/835/play#15014' 'http://bangumi.bilibili.com/anime/3462/play#89073' 'http://bangumi.bilibili.com/anime/1660/play'
Site:       bilibili.com
Title:      我们仍未知道那天所看见的花的名字。 [1 超和平busters]
Type:       Flash video (video/x-flv)
Size:       176.81 MiB (185402909 Bytes)

Site:       bilibili.com
Title:      双星之阴阳师 [17 师傅赋予的红色证明]
Type:       Flash video (video/x-flv)
Size:       259.48 MiB (272085344 Bytes)

Site:       bilibili.com
Title:      夏目友人帐 [1 猫和友人帐]
Type:       Flash video (video/x-flv)
Size:       143.16 MiB (150117218 Bytes)
```
部分修正了 #1735

3. 把timeout的try except block限定到download_urls上去，修正#1897
**timeout增加到15**
```
./you-get http://www.bilibili.com/video/av9954010/
Site:       bilibili.com
Title:      【西瓜新】一个小小的八音盒【有一点花絮】
Type:       MPEG-4 video (video/mp4)
Size:       25.6 MiB (26847580 Bytes)

Downloading 【西瓜新】一个小小的八音盒【有一点花絮】.mp4 ...
 100% ( 25.6/ 25.6MB) ├█████████████████████████████████████████┤[1/1]    9 MB/s

Downloading 【西瓜新】一个小小的八音盒【有一点花絮】.cmt.xml ...
```

```
 ./you-get http://bangumi.bilibili.com/anime/1660/play
Site:       bilibili.com
Title:      夏目友人帐 [1 猫和友人帐]
Type:       Flash video (video/x-flv)
Size:       143.16 MiB (150117218 Bytes)

Downloading 夏目友人帐 [1 猫和友人帐].flv ...
[...]

PR #2026, opened 2017-05-25
[bilibili] use faker to fix 403 problem (CDN rejecting Python urllib's user agent)

zmwangx:
Bilibili's CDN now may reject requests with 403 when `urllib`'s default user agent is used. This appears to be location-based (and started today for me; others might have experienced this for a while): I tested on a bunch of U.S.-based nodes and all requests were rejected without a spoofed user agent, but when I tested on a Mainland China-based node, the requests went through.

To actually use the faker I had to make changes to faker logic in `common.py` (currently when `faker` is `True`, `headers` is simply ignored). Please refer to the commit message for details.

This PR appears to overlap with #1698, which attempts to do a bunch of things and hence hard for me to quickly evaluate. It also seems to roll its own user agent instead of using the existing faker. This PR only attempts to do one thing and hopefully do it well.

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -308,7 +308,7 @@ def urlopen_with_retry(*args, **kwargs):
         except socket.timeout:
             logging.debug('request attempt %s timeout' % str(i + 1))
 
-def get_content(url, headers={}, decoded=True):
+def get_content(url, headers=None, decoded=True):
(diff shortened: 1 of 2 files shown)


soimort:
#1698 has been merged since it fixes a few more long-standing problems. Thanks anyway!

PR #2095, opened 2017-06-19
fix miaopai mobile page download

yunair:
miao pai mobile page url does not contains .htm, so fix it when get scid


soimort:
Thanks :+1:

PR #2121, opened 2017-07-02
fix 'reference before declaration' error in youtube extractor

YenvY:
just add a forward declaration of local variable `ytplayer_config` in Youtube.prepare method since the variable is invisiable to the expression in line 202.


soimort:
Any sample URL for testing this issue? Thanks.

soimort:
Never mind, I found that any content-warning video is sufficient to raise this issue.

Thanks for the fix! :+1:

PR #2138, opened 2017-07-09
output refer and ua message in json

wwqgtxx:


coslyk:
Not all VideoExtractors has attribute "refer" and "ua"
So they should be put in a "try...except AttributeError" block.

wwqgtxx:
@coslyk please lookup
https://github.com/soimort/you-get/blob/develop/src/you_get/extractor.py#L33
and
https://github.com/soimort/you-get/blob/develop/src/you_get/extractor.py#L34
it let all VideoExtractor has those attribute

rosynirvana:
The ```VideoExtractor``` pass to ```output``` is not necessarily a real VideoExtractor from extractor.py. It can also be a dummy one, defined in output_json.py

It's tricky and should be merged into common.py.

wwqgtxx:
@rosynirvana This is a design problem from @sceext 's
https://github.com/wwqgtxx/you-get/commit/1d7758b1070442b574a89ae049b94581936a2e41
But it was in the you-get code for about two years, and I don't want to make a big change for this module.
This PR could work fine with most `VideoExtractor`, if you find which `VideoExtractor` will cause the `AttributeError` except the output_json.py's `VideoExtractor`, please tell me,and then I will add the try except code to avoid this problem.

PR #2189, opened 2017-07-24
[instagram] Fix always downloading to current dir, ignoring `-o`

kolen:
Fix Instagram downloader always downloading to current directory, ignoring `-o` option.

#2188


soimort:
Nice catch. Thank you!

PR #2319, opened 2017-09-03
[common] retry if response.read() timeout

JayXon:
Recently I started to get timeout in response.read(), this commit will make it retry automatically if timeout.


soimort:
Looks good. Thanks!

PR #2425, opened 2017-10-11
twitter download fix case https://twitter.com/Newstv_jp/status/914293814054891521

Cphilo:


soimort:
You need to be more careful, since dropping `ext_tw_video` will break URLs like: https://twitter.com/CBSNews/status/918810067985723393
better to capture both strings in regex.

Anyway, thanks for the PR!

PR #2484, opened 2017-11-06
qq video does not work well any more

jhc888007:
I don't know why. But downloading qq video seemed not worked well. I delete some lines. Then it works.


soimort:
Thanks!

PR #2490, opened 2017-11-12
pep8 update for common.py

iawia002:
In order to better read and understand the code, I updated the common.py based on the pep8 guideline(no logic change)

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -1,5 +1,25 @@
 #!/usr/bin/env python
 
+import os
+import re
+import sys
+import time
+import json
+import socket
+import locale
+import logging
+import argparse
+from http import cookiejar
+from importlib import import_module
+from urllib import request, parse, error
+
+from .version import __version__
+from .util import log, term
+from .util.git import get_version
+from .util.strings import get_filename, unescape_html
+from . import json_output as json_output_
+
+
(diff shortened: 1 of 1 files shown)


iawia002:
Another thing that puzzles me is the format of SITES.

```python
SITES = {
    '163'              : 'netease',
    '56'               : 'w56',
    'acfun'            : 'acfun',
    'archive'          : 'archive',
    'baidu'            : 'baidu',
}
```

I mean it 'looks good', but I don't think it's standard python dictionary format.

how about the format below 🤔

```python
SITES = {
    '163': 'netease',
    '56': 'w56',
    'acfun': 'acfun',
    'archive': 'archive',
    'baidu': 'baidu',
}
```

iawia002:
Another question is the format of function with too many parameters, there are two formats, I don't know which one is cooler

```python
def download_url_ffmpeg(
    url, title, ext, params={}, total_size=0, output_dir='.', refer=None,
    merge=True, faker=False, stream=True
):
    pass
```

```python
def download_url_ffmpeg(
        url, title, ext, params={}, total_size=0, output_dir='.', refer=None,
        merge=True, faker=False, stream=True):
    pass
```

soimort:
Very good, thanks!

PR #2498, opened 2017-11-18
[vine] fix empty file name if no description found

zmux:
Now if no description in video file metadata found then user name and
video id will be used


soimort:
Thanks for the fix!

PR #2502, opened 2017-11-19
Add coub support

zmux:
Video and audio files with high quality are downloaded.
We use ffmpeg for audio and video combining into one file with looping support


cnbeining:
Hi,

Thanks for your PR.

Can you kindly rebase all the commits into 1 or 2 commits? 

Thanks,

PR #2503, opened 2017-11-20
Update the regex for mgtv extractor, adapt to https

chinsyo:
Mgtv extractor failed with `https` link, this pull request has fix the problem. Now it works on my mac both `http` and `https` links.


chinsyo:
Relate to #2452 , accept this pull request will solve the problem.

soimort:
Thanks for the fix.

PR #2525, opened 2017-11-27
[coub] add coub.com support

zmux:
Video and audio files with high quality are downloaded.
We use ffmpeg for audio and video combining into one file


soimort:
Good. Thanks!

zmux:
No problem, glad I can help!

PR #2533, opened 2017-12-07
[bilibili] fix title regex

iawia002:
bilibili got a weird page([https://www.bilibili.com/video/av16907446/](https://www.bilibili.com/video/av16907446/)), there is no `=` between the `title` property and it's value.

So I changed the regular expression, now it directly matches the contents of the `h1`.

I also updated the unit test.

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -125,7 +125,7 @@ class Bilibili(VideoExtractor):
         self.referer = self.url
         self.page = get_content(self.url)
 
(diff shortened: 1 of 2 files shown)


soimort:
Could you please split it into two separate commits (one for Bilibili fix and one for unit test formatting)? Thanks.

iawia002:
sure, I will do it tonight.

iawia002:
done for separating, ping @soimort 
good to merge now?

soimort:
:+1:

PR #2534, opened 2017-12-08
[youku]fix ccode

Amehinou:
fix the ccode


soimort:
What's the point of this fix?

For now `0590` does not yield 1080P at all. Both `0508` and `0512` would do.

PR #2541, opened 2017-12-15
[kuaishou] add missing intro

iawia002:


iawia002:
It seems that some of the tests failed due to the network reason.

soimort:
Unfortunately, tests like this rely too much on the network status (the availability of service providers; the connectivity between US and mainland China, where I assume its server is located in; also the availability of videos used as part of tests, for they could be taken down at any time). It can break the Travis CI very easily.

Generally I prefer not to add more cases for testing actual video URLs. In this case, appending the site name to the `README` list should be sufficient.

iawia002:
@soimort updated!
I agree with you, but I still think that we should think a way to test whether these download functions are available. Waiting for the user to report the error is too inefficient.

soimort:
:+1:

PR #2553, opened 2018-01-04
[bilibili] fix 403 error when requesting interface.bilibili.com

zmwangx:
interface.bilibili.com now requires referer to be set.


zmwangx:
Actually, this only fixes `--info`, `--url` and `--json`; actually downloading the URLs result in HTTP 466. It is not immediately obvious to me how to fix that part; I know that downloading from `tx.acgvideo.com` now requires a valid user agent in addition to referer, but the URLs extracted by you-get can't be downloaded either way. I'll look into it later.

zmwangx:
Closing this since #2555 should work.

PR #2556, opened 2018-01-10
Update youku.py

9yw:
ccode=0512会报错 客户端无权播放,201 改成0513可以下载
got error 201 when ccode=0512

PR #2557, opened 2018-01-11
change ccode into 0502,Then youku will work

elespec:
change ccode into 0502,Then youku will work


laoniustudio:
when I download you ku video, it works fine at beginning, but shows forbidden error at the progress about 47% or 53% . how to do with it, thanks

9yw:
0502也不行了，看 #2556 ，改0513可以。

elespec:
Youku api developer change their api so frequently,It looks they really like that,So sad;

PR #2576, opened 2018-02-21
Added an auto rename option and fixed the force option

John128217:
- [x] Add -a to your you-get code to auto-rename your file.
- [x] Fixed -f (force)

I've noticed that if I download two videos with same names from youtube (e.g. https://www.youtube.com/watch?v=606hmlA_nxw and https://www.youtube.com/watch?v=CLrXTnggUeY), 
only one of them will be saved (usually the bigger one according to the original script `os.path.getsize(output_filepath) >= total_size * 0.9`). However, I want them both while preserving their names from youtube. It looks like there are a lot of changes, but I just added an indent and everything changed.
Also, I've noticed that "force" is not working at all. I fixed that issue.

I'm quite new to coding... Any feedback is appreciated!

not working -f (old version):
not working force (old version)

fixed -f : (-a does not matter here)
fixed force

without this new option:
without this new option

with this new option:
with this new option

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -622,10 +622,13 @@ def url_save(
                         bar.done()
                     if not force and auto_rename:
                         path, ext = os.path.basename(filepath).rsplit('.', 1)
-                        if (re.compile(' \(\d\)').match(path[-4:]) is None):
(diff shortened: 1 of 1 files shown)


soimort:
Looks good, thanks!

PR #2581, opened 2018-03-10
Use 0513 when cookies are used

Kugel-Blitz:
0507 doesn't seem to honour cookies when they're loaded, whilst 0507 worked just fine.


soimort:
:+1:

PR #2592, opened 2018-03-29
[youku] use default ckey

JayXon:
1080p works


luoyimin925:
Thank you very much! It works now.

soimort:
:+1:

PR #2593, opened 2018-04-22
fix bilibili format

ljhcage:
update xml-url and sec1 to get all format of bilibili-video


soimort:
Thanks for the fix! :+1:

PR #2595, opened 2018-05-11
[qq] fix some error cases("check vid&filename failed" and "format inv…

LittleRed-YXC:
i got some error cases when downloading qq videos, the msg that qq return is "check vid&filename failed".


soimort:
Thanks :+1:

PR #2598, opened 2018-05-17
new bangumi api url

fz19870823:
bilibili has a new bangumi api url for paid members.
**http://www.biliplus.com/BPplayurl.php?**


fz19870823:
sorry that was changed by a chrome plugin................

PR #2145, opened 2017-07-12
Add flake8 to testing

cclauss:

PR #2599, opened 2018-05-18
Add flake8 to the testing (again)

cclauss:
Another attempt at #2145

Output: https://travis-ci.org/soimort/you-get/jobs/380561806#L453-L458

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ python:
   - "3.6"
   - "nightly"
   - "pypy3"
+before_install: pip install flake8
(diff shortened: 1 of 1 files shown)


soimort:
While PEP-8 conformance would be a nice thing, adding flake8 to tests without fixing existing code will make Travis CI break forever thus become less helpful on detecting other (probably more significant) issues.

I suppose this should be merged at a later point. (mini-PRs correcting coding style are welcome, too.)

soimort:
Thanks for the PR.

`flake8` doesn't seem to support Python 3.2, skipping that version.

PR #2651, opened 2018-11-01
use new bilibili live api, fix None Content-Type

lc4t:
Use new API: `https://api.live.bilibili.com/room/v1/Room/playUrl?cid={}&quality=0&platform=web` to get .flv url

And sometimes, response.headers['Content-Type'] is None, fix the `None` to `''`


soimort:
Nice. Thanks! :+1:

PR #2657, opened 2018-11-16
Fix bilibili title regex

bitdust:
old regex can only aim at this:
48576585-e21c6500-e94f-11e8-8bae-1659a66ef758

this PR match the title pattern with html attribute like this
48576587-e2b4fb80-e94f-11e8-8fb9-0758e8e32275

PLZ test the code using `--cookies` to login Bilbili and fetch the 2nd kind of title pattern

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -130,7 +130,7 @@ class Bilibili(VideoExtractor):
         m = re.search(r'<h1.*?>(.*?)</h1>', self.page) or re.search(r'<h1 title="([^"]+)">', self.page)
         if m is not None:
             self.title = m.group(1)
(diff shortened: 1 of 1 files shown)


soimort:
Could you please provide the URL of a sample video?

Also, it would be a good idea to squash all your commits into one.

bitdust:
The test URL is : https://www.bilibili.com/video/av12828695
different title pattern occurs when using `--cookies` to assign the login cookie.
As for the squash problem, may i close this PR and pull a new one?

soimort:
You may use `git reset --soft HEAD~3` and commit again. Or you can just create a new PR if you prefer.

bitdust:
thank your for your instruction.

PR #2659, opened 2018-11-20
fix bilibili title regex

bitdust:
old regex can only aim at this:
48576585-e21c6500-e94f-11e8-8bae-1659a66ef758

this PR match the title pattern with html attribute like this
48576587-e2b4fb80-e94f-11e8-8fb9-0758e8e32275

TEST URL:
https://www.bilibili.com/video/av12828695

different title pattern occurs when using `--cookies` to assign the login cookie.
PLZ test the code using `--cookies` to login bilbili and test the 2nd type of title pattern.


soimort:
:ok_hand:

PR #2660, opened 2018-11-22
fix acfun bangumi page

FengLi666:
url pattern and html layout of acfun's bangumi page is different from other acfun page , which causes problem
this pr fixs it

~~~
you-get -i http://www.acfun.cn/bangumi/ab5022161
Site:       AcFun.tv
Title:      但是不屈不挠 SAGA (acfun)
Size:       539.6 MiB (565815129 Bytes)
~~~


soimort:
:+1:

PR #2665, opened 2018-12-05
fix ixigua downloading failure

PerrorOne:
The ixigua download will result in such an error:
ep7a5x k0odmp9dut3 rat
Now it can start working:
v fd 21 vyqa 6s_h 5 f
playlist:
tkdv gsasr0 c yoc3dch


soimort:
:+1:

PR #2168, opened 2017-07-20
[youtube]use mp4_audio track when no audio track for webm

rosynirvana:
#2167 

the reason for UnboundLocalError for dash_webm_a_url is there could be no audio track with ```audio/webm```

```
video/webm;+codecs="vp9"
video/mp4;+codecs="avc1.640032"
video/webm;+codecs="vp9"
video/mp4;+codecs="avc1.640028"
video/webm;+codecs="vp9"
video/mp4;+codecs="avc1.4d401f"
video/webm;+codecs="vp9"
video/mp4;+codecs="avc1.4d401e"
video/webm;+codecs="vp9"
video/mp4;+codecs="avc1.4d401e"
video/webm;+codecs="vp9"
video/mp4;+codecs="avc1.4d4015"
video/webm;+codecs="vp9"
video/mp4;+codecs="avc1.4d400c"
video/webm;+codecs="vp9"
audio/mp4;+codecs="mp4a.40.2"
```

16 tracks, 15 for videos and only one for audio
I believe the correct way is to fetch the ```audio/mp4``` with the ```video/webm``` track

```
./you-get -id https://www.youtube.com/watch\?v\=VaoAU6PvuTg
[DEBUG] get_content: https://www.youtube.com/get_video_info?video_id=VaoAU6PvuTg
[DEBUG] get_content: https://www.youtube.com/watch?v=VaoAU6PvuTg
[DEBUG] get_content: https://www.youtube.com/yts/jsbin/player-vflMaap-E/ja_JP/base.js
site:                YouTube
title:               青青河边草，最后的绿地之旅
streams:             # Available quality and codecs
    [ DASH ] ____________________________________
    - itag:          313
      container:     webm
      quality:       3840x2160
      size:          1146.6 MiB (1202279479 bytes)
    # download-with: you-get --itag=313 [URL]

    - itag:          264
      container:     mp4
      quality:       2560x1440
      size:          635.0 MiB (665812113 bytes)
    # download-with: you-get --itag=264 [URL]

    - itag:          271
      container:     webm
      quality:       2560x1440
      size:          564.1 MiB (591452544 bytes)
    # download-with: you-get --itag=271 [URL]

    - itag:          137
      container:     mp4
      quality:       1920x1080
      size:          271.5 MiB (284663035 bytes)
    # download-with: you-get --itag=137 [URL]

    - itag:          248
      container:     webm
      quality:       1920x1080
      size:          187.4 MiB (196451593 bytes)
[...]

PR #2676, opened 2019-01-21
Use copy when merging audio and video

wangqr:
Currently `you-get` is re-encoding the audio using aac when merging audio and video tracks. This is a waste of time and will degrade the quality of audio. This PR changes the codec for audio to copy.

Further more, the `-strict experimental` option is only necessary when using the aac codec and has already been deprecated in 2015. This PR also removes this option.

### DIFF
--- a/src/you_get/processor/ffmpeg.py
+++ b/src/you_get/processor/ffmpeg.py
@@ -59,12 +59,7 @@ def ffmpeg_concat_av(files, output, ext):
     params = [FFMPEG] + LOGLEVEL
     for file in files:
         if os.path.isfile(file): params.extend(['-i', file])
-    params.extend(['-c:v', 'copy'])
-    if ext == 'mp4':
-        params.extend(['-c:a', 'aac'])
-    elif ext == 'webm':
-        params.extend(['-c:a', 'vorbis'])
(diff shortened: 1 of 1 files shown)


soimort:
If I recall correctly, sometimes YouTube can fail to provide a corresponding Vorbis audio track to pair with its video track when getting a DASH stream with WebM format (#2168). This usually happens when a video is relatively newly uploaded onto YouTube. That's why we still need to enforce the audio re-encoding into `vorbis` as an AAC audio is used instead. The same goes for MP4 format.

I agree that audio re-encoding isn't very useful when it's already in the right format. Better if re-encoding is only enforced when necessary, and in other cases we may just use `copy`.

As for `-strict experimental`, I'd rather keep it since it works for any version of ffmpeg. Removing it will break compatibility with pre-2015 ffmpeg on older systems, but won't do any good.

wangqr:
I cannot find a example video to reproduce #2168, so I just downloaded both mp4 and webm of a same video without merge, then manually invoke `ffmpeg_concat_av`. Now it checks if the return value of `ffmpeg` is zero and automatically retry.

`opus` is used instead of `vorbis` to be consistent with YouTube.

zwh8800:
The fallback version looks good.

PR #2677, opened 2019-01-25
Use importlib instead of imp in setup.py

wangqr:
Running `setup.py` gives the following output:
```
./setup.py:8: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import os, json, imp
```
This PR replaces the imp module with importlib, according to [this SO answer](https://stackoverflow.com/a/67692/3342190).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,8 @@ PACKAGE_NAME = 'you_get'
 
 PROJ_METADATA = '%s.json' % PROJ_NAME
 
-import os, json, imp
+import os, json
+import importlib.util
 here = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


cclauss (approved):
Looks good to me.

Python 3.4 is EOL now too https://devguide.python.org/devcycle/#end-of-life-branches
* https://devguide.python.org/versions

branchvincent:
Would be great to get this merged, Python 3.12 has removed `imp`: https://docs.python.org/3/whatsnew/3.12.html#imp

pingmoogle (approved):
It works on Python 3.12

xlucn:
Could this be merged? Arch Linux users will suffer from this issue since Python 3.12 is now in the repo.

PR #2678, opened 2019-02-04
fix coub ffmpeg

sheerluck:
Test URL:  https://coub.com/view/19cyub

1) ```params.extend(['-safe', '0'])``` fixes ```[concat @ 0x55b4c8f02980] Unsafe file name 'Test_For_The_Princess!.mp4'```
Before patch: Test_For_The_Princess!_full.mp4 is not created
After patch: Test_For_The_Princess!_full.mp4 created successfully

2) ```round``` fixes ```audio_duration 64.0 + video_duration 8.04 -> 7```
Before patch: video is freezed for last 8 second
After patch: perfect loop


soimort:
Looks good. Thanks!

PR #2683, opened 2019-02-20
cache DASH audio size

ljhcage:
in some condition, query dash audio size too many time can cause HTTP error 907, for example https://www.bilibili.com/video/av39021966. And cache audio size can avoid query same thing again, especially in audio size


soimort:
Merged. Thanks!

PR #2685, opened 2019-02-26
Bilibili - Failed to download all videos from one channel

ooooo999:
In the case when there are more than 100 videos uploaded by one Bilibili user, running "you-get [URL]" or "you-get -l [URL]" will only download the latest 100 videos.

Debug shows that the refactor turns to url (https://space.bilibili.com/ajax/member/getSubmitVideos?mid=19111402&page=1&pagesize=100&order=0&jsonp=jsonp) .
Looked into source codes. The value of page=1 and pagesize=100 are somewhat hard-coded. Tried changing pagesize to 200 returns an error. 
Possible approach is 1)to somehow fix the error and allow it to download all videos from the up-loader no matter how many they are, or 2)add a command line option/parameter for user to specify the page that he's trying to start downloading from.

Test case:
you-get https://space.bilibili.com/19111402/video
you-get -l https://space.bilibili.com/19111402/video

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -7,11 +7,9 @@ from you_get.extractors import (
 )
 
 
-class YouGetTests(unittest.TestCase):	
-    def test_bilibili(self):
-
-        bilibili.download('https://space.bilibili.com/19111402/video', info_only=True)
-        youtube.download('https://space.bilibili.com/19111402/video', info_only=True)
+class YouGetTests(unittest.TestCase):
+        def test_bilibili(self):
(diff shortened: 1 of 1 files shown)


sovereign000:
you-get https://space.bilibili.com/479372369/video -l  can not work!

File "k:\software\python38\lib\site-packages\you_get\extractors\bilibili.py", line 737, in download_playlist_by_url
    pc = videos_info['data']['pages']
TypeError: string indices must be integers

qchenac:
This problem still not addressed yet. Really need to expand the limit to 200 or something.

PR #2695, opened 2019-04-17
Travis CI: Add flake8 test F632 use ==/!= to compare str, bytes, and int literals

cclauss:
Also:
* Upgrade to Python 3.7 production instead of pre-release
* Upgrade to current Python nightly and 3.8-dev builds
* Remove __sudo__ tag which is now deprecated on Travis CI

[__Python < 3.5 are all end-of-life.__](https://devguide.python.org/devcycle/#end-of-life-branches)  At what point do we drop support for them?

$ __flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics__
```
./src/you_get/extractors/baidu.py:41:20: F632 use ==/!= to compare str, bytes, and int literals
    return None if lrc is '' else "http://music.baidu.com%s" % lrc
                   ^
./src/you_get/extractors/zhibo.py:37:8: F632 use ==/!= to compare str, bytes, and int literals
    if is_live is not "1":
       ^
./src/you_get/extractors/douyutv.py:65:8: F632 use ==/!= to compare str, bytes, and int literals
    if server_status is not 0:
       ^
./src/you_get/extractors/douyutv.py:70:8: F632 use ==/!= to compare str, bytes, and int literals
    if show_status is not "1":
       ^
./src/you_get/extractors/panda.py:28:8: F632 use ==/!= to compare str, bytes, and int literals
    if status is not "2":
       ^
5     F632 use ==/!= to compare str, bytes, and int literals
5
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,15 +6,20 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  - "3.7-dev"
-  - "nightly"
   - "pypy3"
+matrix:
+  include:
+    - python: "3.7"
+      dist: xenial
+    - python: "3.8-dev"
+      dist: xenial    
+    - python: "nightly"
+      dist: xenial    
 before_install:
   - if [[ $TRAVIS_PYTHON_VERSION != '3.2'* && $TRAVIS_PYTHON_VERSION != '3.3'* ]]; then pip install flake8; fi
 before_script:
-  - if [[ $TRAVIS_PYTHON_VERSION != '3.2'* && $TRAVIS_PYTHON_VERSION != '3.3'* ]]; then flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics; fi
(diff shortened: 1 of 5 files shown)


soimort:
Thanks a lot for the improvement.

> Python < 3.5 are all end-of-life. At what point do we drop support for them?

Good reasons for moving forward would be advantageous libraries such as `asyncio` and `typing`, and 3rd party ones like Requests. Since 2016 I'd been thinking about a complete rewrite of you-get based on these, with more structured code and a more extensible interface, but given my limited time, I'm sorry to say it may not happen anytime soon.

I tend not to break compatibility with legacy Python just because of the syntactic sugar or handy function parameters that newer versions introduced. So I guess the overall codebase will mostly stay in statu quo until I find more time to work out the rewriting plan, in which support for legacy Pythons will eventually be dumped (due to the use of new facilities like type hints etc.).

PR #2696, opened 2019-04-17
Fix iQiyi broken api blocking video download

dbw9580:
The API `http://mixer.video.iqiyi.com/jp/mixin/videos/` used by the iQiyi extractor to get correct video title seems to have been deprecated and is returning 404 error. For example, for this video `https://www.iqiyi.com/w_19s2jsr8z1.html`, whose `tvid` is `27529587009`, requesting `http://mixer.video.iqiyi.com/jp/mixin/videos/27529587009` returns

```
var tvInfoJs=<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

which is not valid JSON.

This causes the JSON parser to fail and throws an unexpected exception that prevents the download from starting. <https://github.com/soimort/you-get/blob/2e461820f8dbb46bf7aa973e9df02986e758dd10/src/you_get/extractors/iqiyi.py#L141>

Fortunately, the video downloading part is working perfectly fine. I managed to find a new API that is returning video title and other stuff working as expected. Simply replacing the API fixes the problem.

I have manually tested the new API with a few videos from iQiyi's front page, and it seems to be working fine. However, I am not familiar with contribution conventions in this project, so if there are any other things that needs to be done on my part, please let me know.

### DIFF
--- a/src/you_get/extractors/iqiyi.py
+++ b/src/you_get/extractors/iqiyi.py
@@ -136,12 +136,9 @@ class Iqiyi(VideoExtractor):
                       r1(r'vid=([^&]+)', self.url) or \
                       r1(r'data-player-videoid="([^"]+)"', html) or r1(r'vid=(.+?)\&', html) or r1(r'param\[\'vid\'\]\s*=\s*"(.+?)"', html)
             self.vid = (tvid, videoid)
-            info_u = 'http://mixer.video.iqiyi.com/jp/mixin/videos/' + tvid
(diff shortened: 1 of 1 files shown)


soimort:
Very good. Thanks!

Tobyzgj:
Youku 201 does not have the right to play how to solve the problem

PR #2698, opened 2019-04-22
add new video type for bilibili.com

WingGao:
fix the new type of `高清 1080P+` on bilibili.com
test command:
```bash
# you need a vip account
you-get --debug -c bilibili.txt --playlist -i "https://www.bilibili.com/video/av40115478"
```

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -12,7 +12,8 @@ class Bilibili(VideoExtractor):
     stream_types = [
         {'id': 'flv_p60', 'quality': 116, 'audio_quality': 30280,
          'container': 'FLV', 'video_resolution': '1080p', 'desc': '高清 1080P60'},
-        # 'id': 'hdflv2', 'quality': 112?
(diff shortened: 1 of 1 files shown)


src/you_get/extractors/bilibili.py
```diff
@@ -111,6 +112,13 @@ def bilibili_space_video_api(mid, pn=1, ps=100):
     def bilibili_vc_api(video_id):
         return 'https://api.vc.bilibili.com/clip/v1/video/detail?video_id=%s' % video_id
 
+    @staticmethod
+    def url_size(url, faker=False, headers={},err_value=0):
+        try:
```
soimort: Why is this wrapper function necessary? Under what circumstances can `url_size` from common.py fail?

WingGao: Because sometime the server will return a `403` error for the audio file, but the video file is accessible.

Tobyzgj:
Youku 201 does not have the right to play how to solve the problem？

soimort:
:+1:

PR #2699, opened 2019-04-26
Fix bilibili vip user bangumi download error

FSpark:
I found that the difference in `initial_state`  between vip and non-vip when downloading bangumi caused many errors and then fixed it.  By the way, 1080P+ bangumi support has been added.
Test command(**need a vip account**):
`you-get 'https://www.bilibili.com/bangumi/play/ep250453' -i -c bilibili.txt`
`you-get 'https://www.bilibili.com/bangumi/play/ep250453' -i `

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -281,15 +281,27 @@ class Bilibili(VideoExtractor):
                 log.w('This bangumi currently has %s videos. (use --playlist to download all videos.)' % epn)
 
             # set video title
-            self.title = initial_state['h1Title']
+            # need a better way to judge whether it is a vip
+            if 'epPayMent' in initial_state:
+                is_vip = False
+            elif 'epStat' in initial_state:                
+                is_vip = True
+            else:
+                is_vip = False
+
+            if is_vip :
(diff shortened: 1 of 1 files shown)


src/you_get/extractors/bilibili.py
```diff
@@ -281,15 +281,27 @@ def prepare(self, **kwargs):
...
             # set video title
-            self.title = initial_state['h1Title']
+            # need a better way to judge whether it is a vip
+            if 'epPayMent' in initial_state:
+                is_vip = False
+            elif 'epStat' in initial_state:                
+                is_vip = True
+            else:
+                is_vip = False
+
```
soimort: This is logically equivalent to
```py
is_vip = 'epStat' in initial_state and 'epPayMent' not in initial_state
```
but why is it not a vip when 'epPayMent' exists? (just curious, I don't have a VIP account to test myself)

FSpark: The root cause is the difference between the new version and the old version of bilibili player. The player of the previous bilibili bangumi page is the old version by default, but after just trying again, I found that there is no difference now, because it seems to have changed now. However, this doesn't mean that the current you-get can work normally, HD quality is still not available, I will fix it again soon.

PR #2713, opened 2019-06-18
Fix the problem that kuwomusic can't be download

licabbage:
Change the current kuwo url to a previous version.
(example: change"http://www.kuwo.cn/play_detail/7149583" to "http://www.kuwo.cn/yinyue/7149583")


soimort:
Please give your code a test and see if it really works (for the example URL you provided as above).

PR #2715, opened 2019-07-02
Fix the parser issue for acfun videos.

xingheng:
The original html attributes have changed since the last acfun site's upgrade, it fails to extract the video id and title from the serial `data-xxx` attributes, this pr fixes it by extracting the metadata info from the json data instead.

Please help to review this.


soimort:
Thanks :+1:

PR #2716, opened 2019-07-08
A request

L1cardo:
I have a small request:
  Can we just download videos like this , for example    `you-get https://www.youtube.com`
  Instead of `you-get 'https://www.youtube.com'`
Can we remove the `' ' `here ?
Because when I want to download a video , I would copy from browser and just paste behind `you-get`, and it would not be convenient if I add `' '`
Thanks


soimort:
You may leave out the single quotes if you want. But you'll need to escape special characters like `?` as `\?` so that they won't be consumed by shell.

This is basically how shell works, and there's nothing I can do about it.

PR #2719, opened 2019-07-16
changed youku ccode to 0519

yaotianran:
changed youku ccode to 0519


soimort:
Thanks :ok_hand:

kevinsingapore:
change the ccode to 0519 not ok；
errors like the this:
image

kevinsingapore:
> change the ccode to 0519 not ok；
> errors like the this:
> image

i have the vip id and got the cookie，but it's not ok;

you-get: version 0.4.1355

PR #2724, opened 2019-08-05
ss18/pr0 trivial

ss18:
Fixed typos:

avalable -> available
basicly -> basically
infomation -> information
Unkown -> Unknown

Found with https://github.com/ss18/grep-typos


soimort:
Good catch:sweat_smile:

Thanks!

PR #2733, opened 2019-08-28
modify encoding with open cookies file

Shaun-Chen:
fix a bug for reading cookies.txt：the UnicodeDecodeError, if cookies.txt include Chinese characters
Example：
cookies.txt include：.v.qq.com       TRUE    /       FALSE   0       ptag    www_baidu_com|movie_v3_new:img:扫毒2天地对决。
The Error：UnicodeDecodeError: 'ascii' codec can't decode byte 0xe6 in position 4935: ordinal not in range(128)


soimort:
Thanks!

PR #2751, opened 2019-11-14
Fix bug on flv only playlist video

ljhcage:
1. Able to download video from Pth video selected by url from a muti-part video
2. Fix bug on flv only playlist video, sort stream is necessary for it


soimort:
Thanks :+1:

PR #2753, opened 2019-11-17
fix extractors\naver.py

SFMDI:
The recent video on tv.naver was not downloaded, but only the wrong thumbnails were downloaded in the previous code. So I solved this problem by integrating the if-else statement into one.

It also resolved the problem of not being properly stored in the specified output_dir.

* example of wrong result
result of "you-get https://tv.naver.com/v/10984456"
only downloaded thumbnails no video
thumnail

* can download this video correctly after code change
correct


SFMDI:
I made the mistake of putting another pool request here.. I'll close this pool request and reopen it. sorry for that

PR #2759, opened 2019-12-01
Fix playback of VideoExtractor.dash_streams

isjerryxiao:
- Playing dash_streams is impossible since launch_player is expecting `List[str]`, while VideoExtractor.dash_streams has a type of `List[List[str]]`.  
- This patch should fix that problem and allow dash_streams to be played.


soimort:
Thanks for the PR!

PR #2761, opened 2019-12-14
Add an audio extraction from downloaded video option

SFMDI:
- option -e  :  you can extract audio from downloaded video

- option -E FORMAT :  set extacted audio format. default format is mp3

When I downloaded a video for the purpose of listening to music, I felt burdened by the additional capacity that would result from receiving the video itself other than the music.
So using one FFmpeg, which is the Prerequisites of this package, I added the option to extract sound only from downloaded videos.

-E option can determine the extension of sound extracted, but does not guarantee proper operation unless it is a sound file extension.

If the file you are trying to extract sound from is a sound file, for example, a sound file with a different extension is created, for example, from a flav file to an mp3 file.

If you want to extract sound files from an image file, it does not guarantee proper operation.

# example results 

## only option -e
only -e
result -e

## add option -E
add option -E
add option -E result

## extraction file (wrong extension) only downloaded vido
error
error result

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -138,6 +138,8 @@ cookies = None
 output_filename = None
 auto_rename = False
 insecure = False
+extract_audio = False
+extract_format = 'mp3'
 
 fake_headers = {
     'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8',  # noqa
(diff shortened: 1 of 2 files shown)


SFMDI:
I'm using Python 3.6
I don't know why Python 3.4 check failed...

BraveDrXuTF:
do you think you can achieve download audio but not video only?

PR #2764, opened 2019-12-27
fix issue about KeyError: 'url_encoded_fmt_stream_map'

chonpsk:
It looks like working well.
#2755 tested.


chonpsk:
well, I don't know why I edited youtube.py but it failed on imgur.

chonpsk:
Cuz two times failed on different environment but also imgur I think there is something wrong on imgur. Maybe the checker should be changed.

Gelma:
Thanks a lot. Fixed the problem to me also

soimort:
Great job. Thanks!

PR #2770, opened 2020-01-25
Add PySocks as an extra requirement

JarryShaw:
A simple way to make life much easier :)

c.f. `you_get.common.set_socks_proxy`


soimort:
Nice. Thanks!

PR #2772, opened 2020-01-29
fixed coub_data regex

Ok-pass:
Some videos e.g. [https://coub.com/view/286qqx](https://coub.com/view/286qqx) contain `<` characters belonging to tags nested inside
`<script id='coubPageCoubJson' type='text/json'>`.

They get matched early by the regex resulting in an empty variable coub_data.

`Error while downloading files. the JSON object must be str, bytes or bytearray, not NoneType`

Fixed so it matches everything until the closing `</script>` tag.


soimort:
Thanks!

PR #2779, opened 2020-03-14
[baomihua] fix download 403

helong0911:
add request headers
fix urllib.error.HTTPError: HTTP Error 403: Forbidden


soimort:
Thanks! :+1:

PR #2781, opened 2020-03-23
[Bilibili] add support for BVID

brainbush:
ref https://www.bilibili.com/blackboard/activity-BV-PC.html

PR #2782, opened 2020-03-24
[Bilibili] add support for BVID in playlist mode of bilibili

icpz:
Add the missing support for playlist mode.
Thanks https://github.com/soimort/you-get/pull/2781 .


colining:
cool

soimort:
Thanks!

AndroidMark0506:
And we can use the browser's **Developer Tool** - **Console** and code **window.aid** to get the old AV ID.
I recommend updating **the part of download from bilibilibili** based on this.
You can code the implementing, that if the URL is the BV ID, we can get the old AV ID with the above method.

PR #2783, opened 2020-03-24
working for BV and bug patch

Justsoos:
\w+ for "xxxx?p=dddd", \S+ not working


Justsoos:
sth missing, close first

PR #2787, opened 2020-03-29
fix bilibili favlist download

zhufengning:
and updated the api url


soimort:
Thanks!

zhufengning:
There is still a bug. When the download is correctly complete, it will throw an out-of-range error.

PR #2788, opened 2020-03-31
Fails silently on haokan.baidu.com

Gaelan:
Currently, the Baidu downloader simply fails silently when it encounters a URL it doesn't understand. Therefore, this PR both modifies the Baidu downloader to properly generate an error in this case, and adds a test for a particular URL that doesn't work.


ZM-J:
+1
haokan video can't be downloaded via you-get

PR #2809, opened 2020-05-29
KeyError: 'url'

DEC-QI:
#!/usr/bin/env python

import unittest

from you_get.extractors import (
    youtube
)

class YouGetTests(unittest.TestCase):

    def test_youtube(self):
        youtube.download(
            'https://www.youtube.com/watch?v=tOCUE8wRQYY', info_only=True
        )
		
if __name__ == '__main__':
    unittest.main()


soimort:
A similar issue should have been fixed by ##2812. Thanks for reporting!

PR #2814, opened 2020-06-14
[real PR not issue] fixing the problem for some of bilibili videos: best quality lost

mo-han:
It's a real PR with contributing code. Please don't simply close it.
This PR is tested on my side already.
The description and explanation are in the modified code already.
It's just a temporary fix but the cause is found and confirmed, you can do some improvements instead of my code, if you want.

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -217,10 +217,28 @@ class Bilibili(VideoExtractor):
             if playinfo_ is not None:
                 playinfos.append(playinfo_)
             # get alternative formats from API
+            #<new block begin>
+            # direct call the "/x/player/playurl" api with qn=best_quality in advance
+            # for some of bilibili videos, the best quality is missed in the following `for` loop
+            # even with cookies provided, but without this extra code block
+            # I have been suffering this issue for a long time, util now have fixed it myself
+            # example URLs: https://www.bilibili.com/video/BV1Wv411678n, and a lot lot more...
+            # the cause is explained in the following `for` loop code
+            api_url = self.bilibili_api(avid, cid, qn=best_quality)
+            api_content = get_content(api_url, headers=self.bilibili_headers(referer=self.url))
+            api_playinfo = json.loads(api_content)
+            if api_playinfo['code'] == 0:  # success
+                playinfos.append(api_playinfo)
+            else:
+                message = api_playinfo['data']['message']
+            #<new block end>
             for qn in [120, 112, 80, 64, 32, 16]:
                 # automatic format for durl: qn=0
                 # for dash, qn does not matter
                 if current_quality is None or qn < current_quality:
+                    # for some videos, current_quality < best_quality
+                    # in this case, this `if` block simply skip the best quality qn
+                    # then we can only count on another `if` block using "interface" api
                     api_url = self.bilibili_api(avid, cid, qn=qn)
                     api_content = get_content(api_url, headers=self.bilibili_headers(referer=self.url))
                     api_playinfo = json.loads(api_content)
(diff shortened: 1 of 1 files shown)


mo-han:
@Qiuchi0918 
Hi, thanks for the review.
I'm a little bit familiar with python, but very unfamiliar with this project, even just the bilibili module.
The difficulty came from two side: first, this project has messy code (well, but hell mess but still hard to understand, especially those long method such like `prepare_by_url`); second, the web site and API of bilibili is not managed by us, nor controlled by us.
I have made a self-using modified version, it works pretty good, you can try it if you like:
https://github.com/mo-han/mo-han-toolbox/blob/master/mylib/bilibili.py

Qiuchi0918:
@mo-han Ahh thanks.  I was forced to read outstream from you-get in my downloader to acquire source of actual best quality. With your modification, this might no longer be necessary. I’ll give it a try later.
感谢 :D

soimort:
@soimort-bot test this https://www.bilibili.com/video/BV1Wv411678n

soimort:
@mo-han I'm not sure which best quality is missing here (the output from the current `develop` is as shown above). Could you provide a more specific case?

mo-han:
@soimort 
sry, but really couldn't.
it was long before, details and related URLs are faded in memory.
code modification in this commit is not perfect.
maybe it's not good idea to merge this one.

however, still prefer my own [modified version](https://github.com/mo-han/mo-han-toolbox/blob/master/mylib/bilibili.py), which gets rid of the following issues:
- streams sorted by size rather than resolution, while sometimes bilibili produces videos with size of 720p bigger than 1080p (it may be bilibili's fault but i have to change the code)
- sometimes the condition line `if current_quality is None or qn < current_quality:` cause 1080p60 1080p+ streams lost, i don't know why but after bypassing this if code, issue gone

NoNova7:
@mo-han 
Hello, have you  fixed the issue that bilibili api do not return the [1080P+ 高码率] option? I notice that this particularly  happened to some old videos that initially only support 1080P and Bilibili add the 1080P+ option later. The modified one from you can detect [dash-hdflv2 高清 1080P+] with -i but still unable to download. 
Also notice that: the 1080P+ with newest videos have H265 encoding, while the old videos that added 1080P+ later have the H264 encoding. The link you mentioned above is invalid and I cannot access it. 
感谢.

mo-han:
@NoNova7 
the link is out-dated because i changed repo files' names occasionally
the repo path is `mo-han/mo-han-toolbox`
the bilibili lib file path is `mylib/site_bilibili.py`
for now won't edit that link url in last comment because maybe someday it will be invalid again

i simply replaced a "if" line from `if current_quality is None or qn < current_quality:` to `if True:`
which makes VIP quality streams always detectable (didn't notice any fail until now)
since it's detected then it's able to download, unless the target stream is not selected by default
by default you-get choose the "best" stream, which is stream with largest size actually
however sometimes the highest level "quality" stream is not the biggest sized one
i personally met few videos with 720p as the biggest size stream (1080p{60,+} is smaller)
in such cases, you-get sort streams by size from large to small, resulting into lower quality stream sorted on top and downloaded by default
i changed its sort method from "size first" to "quality/resolution/stream_id first"
i hadn't come across aforesaid issues since then (but i'm not 100% sure if this is your case)

you-get choose HEVC stream over AVC ones
i like this feature and didn't touch its code (never dig into that)

there's another bilibili downloader (not developed by me) which is promising:
http://github.com/lifegpc/bili
its author has done great job on parsing almost all kinds of resouces on bilibili site, and succeeds to provide all kinds of downloading features
i don't use it because i don't need it yet

hope this helps

PR #2818, opened 2020-07-15
Use ElementTree instead of deprecated cElementTree.

tirkarthi:
cElementTree has been deprecated for removal and ElementTree should be used in Python 3. Reference : https://bugs.python.org/issue36543


soimort:
Thanks!

tirkarthi:
Thanks @soimort :)

PR #2822, opened 2020-07-24
bilibili ignore SSL problem

hyjipotou2:
120/5000 
When I downloaded Bilibili video using you-get, an SSL error occurred in the program, possibly because Bilibili used a self-signed SSL certificate.

Operating environment: Python3.7 MacOS

Email: hyjipotou2@outlook.com


soimort:
There is already an `--insecure`/`-k` option to allow insecure server connections when using SSL. Unverified SSL context shouldn't be made as the default for everyone, though.

PR #2827, opened 2020-09-07
[Bilibili] fix bilibili 4k

brainbush:
B站的playurl API加了一个参数 fourk=1 时才会输出4K的播放地址，否则不显示。

bilibili has changed their playurl api, we should add fork=1 to the url params or it will not output the 4k url.


soimort:
Thanks :+1:

PR #2828, opened 2020-09-14
fix iqiyi playlist extrator

jseagull:
iqiyi playlist urls changed from `http://www.iqiyi.com` to `//www.iqiyi.com`

extrator is changed to support all 3 kind of link format

1. `http://www.iqiyi.com`
2. `https://www.iqiyi.com`
2. `//www.iqiyi.com`


soimort:
Thanks!

PR #2841, opened 2020-11-06
bilibili xml file is not completed

killsskillss:
xml file is losing every thing after 90mins if the video with danma exists after 90mins

视频时长超过90分钟后的弹幕不会存在于下载的xml文件中

我写的xml排序文件的注释中有详细说明


killsskillss:
may be I have find a way to solve this issue
following by this url
https://github.com/SocialSisterYi/bilibili-API-collect/blob/master/danmaku/danmaku_proto.md
and It's worked
so, thank you
thanks to every one

PR #2846, opened 2020-11-15
fix netease song downloading

Kagamia:
a fake browser header is required for netease (music.163.com)
or it will return `{"msg":"Cheating","code":-460,"message":"Cheating"}` and write a `{title}.none` file.


soimort:
Thanks!

PR #2853, opened 2020-12-15
feat: add lrts extractor

snowyu:


soimort:
Doesn't seem to work for this URL: http://www.lrts.me/book/48

```
$ ./you-get -i 'http://www.lrts.me/book/48' 
Downloading 阿Q正传-鲁迅-无名氏-有声小说下载,懒人听书官网: 0 count ...
```

snowyu:
it only works on free audio(paid or login required  will be ignore).

soimort:
Nice, thanks.

Single-dashed long options are not used by you-get in order to avoid confusion, so I changed them to double dashes only.

PR #2855, opened 2020-12-18
Remove YinYueTai

SuperHuangXu:


soimort:
Thanks for the info. We should remove everything about YinYueTai (e.g. [`src/you_get/extractors/yinyuetai.py`](https://github.com/soimort/you-get/blob/develop/src/you_get/extractors/yinyuetai.py)) from the repo.

PR #2860, opened 2020-12-26
cannot find 1080p+ format for some videos

NoNova7:
Searched with cookie(b站大会员) enable. Cannot find the "1080P+ 高码率" stream type. Bilibili now adding support about 1080P+高码率 for some old videos that initially only support 1080P, and you-get cannot detect them. But only [some] of videos cannot be detected. (Another example that cannot detect 1080P+: https://www.bilibili.com/video/BV1ns411H7sJ). New videos and some other old videos that have the 1080P+高码率 format nowadays can be detected in you-get.


NoNova7:
modified according to #2814 , yet the tool can now detect the best quality video, but still unable to download it

PR #2859, opened 2020-12-23
mgtv videos download failed

MC-dusk:
芒果TV http://www.mgtv.com 视频下载失败

--debug信息位于

- mgtv - debug1.txt
- mgtv - debug2.txt

PR #2863, opened 2021-01-14
#2859 Mgtv: add another format of url

sinofool:
Also fix indentation causing undefined variable


soimort:
:+1:

shaowei0522:
it works some .I can download all of the pieces of video ,but There was a problem with the last step,when merging into video，All the segments were deleted, but the video was not synthesized

sinofool:
> it works some .I can download all of the pieces of video ,but There was a problem with the last step,when merging into video，All the segments were deleted, but the video was not synthesized

@shaowei0522 Please open another issue with the URL you can't download and debug logs.

PR #2866, opened 2021-01-18
Can not download the youtube subtitles,无法下载油管字幕

CLY88:
升级0.4.15后，无法下载油管字幕，刚开始以为是视频没有字幕，所以没有下载。
后来使用之前已经下载过，且也下载了字幕的视频连接测试，也是没有下载到字幕。
请问字幕下载功能是否能恢复。

After upgrading to 0.4.15, the YouTube subtitles could not be downloaded. At first, I thought that the video did not have subtitles, so I didn't download it.
Later, I have used the video connection I download before for testing, it was downloaded the subtitles, but now it did not download the subtitles also.
Can the subtitle download function be restored?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,4 @@ _*
 
 .vscode
 
+#can not download the youtube captions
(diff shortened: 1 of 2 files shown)


soimort:
Should be fixed in `develop` now. Thanks for reporting!

PR #2886, opened 2021-05-06
skip private video

flewsea:
skip the private video  it cant get urls


soimort:
Merged into [develop](https://github.com/soimort/you-get/commit/144886840212d5d0ee059858e6493dd265927376) instead of master.

PR #2896, opened 2021-07-06
Update douyin.py

pyx:
Updated the extractor for the latest www.douyin.com

### DIFF
--- a/src/you_get/extractors/douyin.py
+++ b/src/you_get/extractors/douyin.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 
 import re
+import json
 from urllib.parse import unquote
 
 from ..common import (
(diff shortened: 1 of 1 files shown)


soimort:
Thanks for the fix!

zsrl:
@pyx 

The update is work ? I get fail result

```
you-get 'https://www.douyin.com/video/6930090637240831239' --debug
```

image

pyx:
> @pyx
> 
> The update is work ? I get fail result
> 
> ```
> you-get 'https://www.douyin.com/video/6930090637240831239' --debug
> ```
> 
> image

Yesterday, right after the new release of you-get 0.4.1536, the website employed some way to prevent clients other than web browser to download the html source, both you-get's downloader and wget will receive 0 (zero) size file while web browser can load the html without problem.

The extractor still work if you-get is able to download the html source (using save page in browser and feed that to you-get, for example), since the fix will probably outside the extractor, I have yet find the time to take a deep look at that.

pyx:
@soimort do you have any idea what might be the cause? I tried with the latest google-chrome UA, and it failed too, so the browser sniffing technique they are using is quite intelligent?

pyx:
For example, `wget 'https://www.douyin.com/video/6982497745948921092'` will save a zero length file, while open https://www.douyin.com/video/6982497745948921092 in browser is okay.

zsrl:
@pyx 
Thanks for answer me. I see the same problem. when I debug by source, variable `page_content` get empty string.

image
image

pyx:
@zsrl It's working now, they probably had back-end bug back then.
yg

zsrl:
@pyx It's really work, thank you very much

wwwfzt:
> @pyx It's really work, thank you very much

is that worked? I see the same problem in version 0.4.1602.
image

PR #2899, opened 2021-07-11
[miaopai] fix weibo videos with h5 api

Kagamia:
match urls like `https://h5.video.weibo.com/show/1034:xxx` and `https://weibo.com/tv/show/1034:xxx`

example:
```
def test_miaopai(self):
    miaopai.download("https://weibo.com/tv/show/1034:4611434789928964?from=old_pc_videoshow", info_only=True)

def test_miaopai2(self):
    miaopai.download("https://h5.video.weibo.com/show/1034:4611434789928964", info_only=True)
```

output:
```
Site:       miaopai
Title:      鬼知道我又看了几遍
Size:       6.36 MiB (6670664 Bytes)
```


soimort:
Nice. Thanks! :+1:

PR #2900, opened 2021-07-15
Update bilibili.py Fix the problem of downloading anime

lihe07:
When downloading anime playlist, "baseUrl" sometimes becomes invalid. This modified version will automatically use "backupUrl" for video and audio downloads.
在下载动漫playlist时，"baseUrl"有时候会失效。此修改版本会自动使用"backupUrl"进行视频及音频下载


SunOfJudge:
Thanks a lot! This does solve the problem when sometimes it fails to download animes.

cfwhj (approved):

PR #2902, opened 2021-07-25
Unable to get data and download video in the same script（无法在同一个脚本中获取数据并下载视频）

zhushenwudi:
Q: Hello, developers, I have encountered a problem, I need use you-get three times to obtain the data. However, problems were found in the following code. After performing getjson and getDefaultDownloadInfo, execute Download, with redirect output, the text exception is obtained, and the download video will be invalid.
问题：你好，开发者，我遇到了一个问题，我需要通过三次调用you-get 方法来获取我想要的数据。但是，在如下代码中发现了问题。执行getJson和getDefaultDownloadInfo之后再执行download，重定向输出后，获取到的文本异常，并且下载视频将无效。

### DIFF
--- a/src/you_get/test.py
+++ b/src/you_get/test.py
@@ -1,0 +1,59 @@
+import traceback
+import sys
+from you_get import common as you_get
+import io
+import os
+
+url = "https://www.bilibili.com/video/BV1GV411p7P9"
+path = os.path.dirname(os.path.realpath(__file__))
+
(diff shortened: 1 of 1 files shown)


zhushenwudi:
lrc 测试 url:
成功的：https://www.bilibili.com/audio/au2033663
失败的：https://www.bilibili.com/audio/au457094

PR #2921, opened 2021-09-16
feat: add first argument support to bilibili

snowyu:
You can specify the first argument(beginning from 0) for bilibili playlist now.

eg: skip the first episode.

```bash
you-get -l --first=1 https://www.bilibili.com/bangumi/play/ss38212
```


xiaowu2001:
I used the command `you-get --playlist --first=8 https://www.bilibili.com/bangumi/play/ss1540` to download 9-13(last episode) of the playlist, but it still downloads from the first episode and gets 13 episodes finally. How can I just download the range of the playlist I want?

snowyu:
This feature has not been mereged via @soimort  .  So you should merge it manually before using.

PR #2935, opened 2021-11-14
Fix zhihu extractor comment: ld

songzy12:
Reference: https://help.doxy.me/en/articles/1160727-video-resolution-change-definition-during-a-call


ljhcage (approved):

soimort:
:+1:

PR #2941, opened 2022-01-07
Add pipenv command usage option

peter279k:
# Changed log

- Adding the `pipenv` command option to let developers can install the `you-get` in the virtual Python environment. And it can let developers separate the environment  from default Python system environment easily :).


soimort:
Thanks! :+1:

PR #2946, opened 2022-02-03
IQiYi Video download failed.

flt6:
When downloading, ffmpeg reports "Failed to open segment 46 of playlist 0", and the reason is "HTTP error 405 Not Allowed" for all of the parts.
As a result, it downloaded nothing.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -3,46 +3,13 @@
 import unittest
 
 from you_get.extractors import (
-    imgur,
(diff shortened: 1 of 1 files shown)


ANYLNK:
I have the same problem

kKevin0001:
俺也一样
temp

20NE:
me too, why close?

PR #2949, opened 2022-02-12
change -safe to 0 for compatibility of ffmpeg5

fakeboboliu:
`-safe -1` set in #1029 in removed in ffmpeg 5.0: https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/46fb395952be32692385449ee214461603480b4b


soimort:
Thanks for the fix! :+1:

PR #2956, opened 2022-04-07
clear stream info

ljhcage:
When download video consecutively, like ```you-get -I download.txt ...```, previous stream info is not clear, may cause download wrong video.
连续下载，例如使用-I模式时，前一个视频的信息未被清除，导致下载视频错误


soimort:
:+1:

PR #2958, opened 2022-05-04
bilibili collectiondetail not supported.

soldatjiang:
dowload [https://space.bilibili.com/364152971/channel/collectiondetail?sid=13909](url) failed


soimort:
Fixed by #2983.

PR #2959, opened 2022-05-04
Updated google search back up to correctly handle non-url parameters

NoahCKnox:
As noted in this [PR](https://github.com/soimort/you-get/pull/2924), the google search back up feature of you-get was not working. 

I have identified and updated the google_search method to correctly handle non-url parameters and correctly return the first youtube video found via the google search of the parameter. Before, the method was causing any non-url param passed into you-get to cause an exception. 

I have also written and added a unit test to the test_common.py file that shows the google search method can correctly return a url when passed in a non-url string.


soimort:
Could you rebase on the latest develop branch?

PR #2979, opened 2022-09-17
Add Python 3.11 release candidate 2 to the testing

cclauss:
Python 3.11 will be on average ___22% faster___ than Python 3.10 so let’s give it a spin...
https://www.python.org/download/pre-releases


soimort:
:+1:

PR #2982, opened 2022-09-22
[netease]: Add pagination on netease radio

StdioA:
`you-get` can only download the first 20 tracks in a radio list, so there's a need to traverse over pages to scrape all the tracks.


soimort:
:+1:

PR #2983, opened 2022-09-26
added support to bilibili collectiondetail api

0x4ea5:
Added support of collectiondetail api of bilibili, tested works with -l option 

`./you-get --format=flv360 -l https://space.bilibili.com/364152971/channel/collectiondetail?sid=13909`

There is already a pull request for the test code: https://github.com/soimort/you-get/pull/2958 @soldatjiang


soimort:
Cool. Thanks:+1:

PR #2986, opened 2022-10-17
[douyin] remove video watermark

sankoshine:
Watermark is removed from videos at both left-top corner and the end of video.


soimort:
:+1:

PR #2992, opened 2023-01-02
Download multipage video collection

arix00:
When there're more than single page videos in a collection, Download all videos as current code will only handle first page.

For 'space_channel_series' and 'space_channel_collection'


soimort:
Nice. Thanks!

PR #2999, opened 2023-02-08
support different codecs for bilibili

URenko:
Support different codecs for bilibili now, especially for AV1, which cannot be fetched before.
See https://github.com/SocialSisterYi/bilibili-API-collect/blob/master/video/videostream_url.md for details.


soimort:
Cool. Thanks :+1:

PR #3000, opened 2023-02-15
feat(bilibili): add support for first option to skip videos

snowyu:
Usage:

Skip the first 10 videos:

```bash
you-get -l --first 10 https://bilibili.com/....
```


super-lulu:
for me the --first param do not work. with this param, the program always download from p1

PR #3001, opened 2023-03-02
fix ixigua download failed

dxsooo:
It seems that the regex should update

### DIFF
--- a/src/you_get/extractors/ixigua.py
+++ b/src/you_get/extractors/ixigua.py
@@ -33,7 +33,7 @@ def ixigua_download(url, output_dir='.', merge=True, info_only=False, stream_id=
         _cookies.append(c.strip().split(' ')[0])
     headers['cookie'] += ' '.join(_cookies)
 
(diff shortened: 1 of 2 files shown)


pein0saga:
> It seems that the regex should update

Still not work. 

`you-get: Get video info from url failed`

src/you_get/extractors/ixigua.py
```diff
@@ -33,7 +33,7 @@ def ixigua_download(url, output_dir='.', merge=True, info_only=False, stream_id=
         _cookies.append(c.strip().split(' ')[0])
     headers['cookie'] += ' '.join(_cookies)
 
-    match_txt = match1(html, r"<script id=\"SSR_HYDRATED_DATA\" nonce=\"argus-csp-token\">window._SSR_HYDRATED_DATA=(.*?)<\/script>")
+    match_txt = match1(html, r"<script id=\"SSR_HYDRATED_DATA\" nonce=.*?>window._SSR_HYDRATED_DATA=(.*?)<\/script>")
```
pein0saga, line 36: It just works ! Thank you !

PR #3002, opened 2023-03-02
Use httpx or reuqests if available 

URenko:
I have cleaned some deprecated request methods.
And then use httpx or reuqests, which has been installed in almost all python environments, if available (they are not required, and will fallback to urllib if neither is present)。

By using httpx/reuqests, we can benefit from:
- Built-in session management
- Built-in decoding (e.g. gzip, brotli)
- Built-in retries
- HTTP/2 if avaliable

I kept them for maximum compatibility with the original processing (urllib).
And has been tested in each of the 3 environments (through `make test` and some personal usages).
If necessary, I can add tests for the 3 environments to GitHub action.

### DIFF
--- a/src/you_get/common.py
+++ b/src/you_get/common.py
@@ -15,6 +15,22 @@ from http import cookiejar
 from importlib import import_module
 from urllib import request, parse, error
 
+try:
+    import httpx
+    session = httpx.Client(transport=httpx.HTTPTransport(retries=3), follow_redirects=True, http2=True,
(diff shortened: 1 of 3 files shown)


URenko:
By the way, I have found a project named "handle all HTTP(s) using Requests" at https://github.com/soimort/you-get/projects?type=classic .
 Do you have plans to deprecate urllib support entirely? If so it could be more elegant, and I'd be glad to help you.

URenko:
Rebased to v0.4.1700

PR #3013, opened 2023-07-03
Download speed limited

Mr-LiuDC:
Watching videos through a proxy in a browser is fast, but downloading through you-get is slow.


savvym:
it's a common case, try [this](https://www.reddit.com/r/youtubedl/comments/11gna52/youtube_is_slow_megathread/) instead.

soimort:
Fixed by #3024.

PR #3014, opened 2023-08-03
[BiliBili] Fix Bangumi (Episode) Download.

RadioNoiseE:
Problem occurred while I was downloading a episode using `you-get`:
IMG_6270
and when I investigate more, I found that the bilibili extracter uses REGEX to try to match the HTML content. Unfortunately, seems like bilibili updated its site and there’s no such element:
IMG_0118
And I think using the content in the HTML is really not a good idea since it may change at anytime. So my solution is to use the API offered by bilibili: `https://api.bilibili.com/pgc/view/web/season`, while it’s return value is
IMG_6271
which includes all the information needed. And itself is robust(stable?).

After my revision, the bangumi (episode) support is fixed:
IMG_6269

(Seems like there are comments like `# FIXME` on the lines that use the REGEX method, is there a need to replace these all with the API method? (I only changed the ones affecting bangumi, since only these are broken))

(Have a nice day)

### DIFF
--- a/src/you_get/extractors/bilibili.py
+++ b/src/you_get/extractors/bilibili.py
@@ -95,6 +95,14 @@ class Bilibili(VideoExtractor):
         return 'https://api.bilibili.com/pgc/player/web/playurl?avid=%s&cid=%s&qn=%s&type=&otype=json&ep_id=%s&fnver=0&fnval=%s' % (avid, cid, qn, ep_id, fnval)
 
     @staticmethod
+    def bilibili_bangumi_ssid_api(season_id):
+        return 'https://api.bilibili.com/pgc/view/web/season?season_id=%s' % season_id
+
+    @staticmethod
+    def bilibili_bangumi_epid_api(ep_id):
+        return 'https://api.bilibili.com/pgc/view/web/season?ep_id=%s' % ep_id
+
+    @staticmethod
     def bilibili_interface_api(cid, qn=0):
         entropy = 'rbMCKn@KuamXWlPMoJGsKcbiJKUfkPF_8dABscJntvqhRSETg'
         appkey, sec = ''.join([chr(ord(i) + 2) for i in entropy[::-1]]).split(':')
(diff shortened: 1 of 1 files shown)


ghtz08:
Tested the link [https://www.bilibili.com/bangumi/play/ss22088](https://www.bilibili.com/bangumi/play/ss22088), can download normally.

clouduser1024:
Please also fix links like "https://www.bilibili.com/cheese"
thousands of thanks!

Whichbfj28:
我也遇到B站下载问题了。因为作者没合并分支。我想问下题主。应该怎么使用你这个更新后的代码。

RadioNoiseE:
> 我也遇到B站下载问题了。因为作者没合并分支。我想问下题主。应该怎么使用你这个更新后的代码。

先从我fork的分支拉取源代码（到当前路径下的`you-get`文件夹）：
```sh
git clone https://github.com/RadioNoiseE/you-get
```
然后用：
```sh
python you-get/setup.py install # append --user to install it locally
```
安装。

如果你系统上有make的话，也可以用：
```sh
cd you-get/
make install
```
来安装（默认为当前用户安装）。

Whichbfj28:
> ```shell
> make install
> ```

感谢。。。

Whichbfj28:
> > 我也遇到B站下载问题了。因为作者没合并分支。我想问下题主。应该怎么使用你这个更新后的代码。
> 
> 先从我fork的分支拉取源代码（到当前路径下的`you-get`文件夹）：
> 
> ```shell
> git clone https://github.com/RadioNoiseE/you-get
> ```
> 
> 然后用：
> 
> ```shell
> python you-get/setup.py install # append --user to install it locally
> ```
> 
> 安装。
> 
> 如果你系统上有make的话，也可以用：
> 
> ```shell
> cd you-get/
> make install
> ```
> 
> 来安装（默认为当前用户安装）。

提示：you-get: You will need login cookies for 720p formats or above. (use --cookies to load cookies.txt.)
提示需要饼干。你是怎么自动化的。到时候cookie过期的话

PR #3019, opened 2024-01-09
fix ffmpeg error: Unrecognized option 'absf'

ifui:
The option '-absf' have change in the new version of ffpmeg.Cannot merge video.
see: [https://github.com/FFmpeg/FFmpeg/blob/master/Changelog](https://github.com/FFmpeg/FFmpeg/blob/master/Changelogl)


soimort:
Nice. Thanks :+1:

PR #3023, opened 2024-05-18
Add workflow test on python 3.12

crnkv:
should pass the checks now... setuptools needs to be upgraded for python 3.12 test to find the module

edit:
still cannot pass unittest... but anyway, failures seemed to be merely network issues...

plus, one may consider migrate from `make test` `python3 setup.py test` unittest to the modern way `pytest tests` (start by having two ways of test in workflow yaml), see [1](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-about-other-commands), [2](https://docs.pytest.org/en/latest/how-to/unittest.html)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


soimort:
Thanks -- The failure was due to some unfixed extractors which are irrelevant to this PR.

PR #3025, opened 2024-06-24
chore: update `MANIFEST.in` to include missing files

chenrui333:
Looks like couple of files are missing in the pypi sdist, so update manifest.in to include them.

relates to https://github.com/Homebrew/homebrew-core/pull/175438


soimort:
Thanks!

PR #3030, opened 2024-09-09
Fix: Restore functionality of -o parameter for Miaopai downloads

doskey:
The -o parameter, used for specifying output filename, was not functioning correctly when downloading videos from Miaopai. This commit addresses the issue and ensures the parameter works as intended.


soimort:
Thanks!

